PR #14, opened 2020-03-26
Fix minor issues

### LINKED ISSUE #11: forgot to import seaborn as sns in metaculus.py
add a seaborn import

### LINKED ISSUE #12: Error if prediction can't actually be submitted
Right now we silently fail

brachbach:
- Fix #11 
- Fix #12 
- Point to correct Johns Hopkins csv

...and test all of these things (some via unit tests, others via integration tests)

### DIFF
--- a/ergo/__init__.py
+++ b/ergo/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '0.2.0'
+__version__ = '0.3.0'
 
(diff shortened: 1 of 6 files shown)


brachbach:
everything is broken without the Hopkins filename fix, so we should merge this ASAP

PR #19, opened 2020-03-27
Dashboardv2

brachbach:
Add and test functionality for dashboard, getting:

1. prediction results and
2. Metaculus questions

as dataframes

### DIFF
--- a/ergo/metaculus.py
+++ b/ergo/metaculus.py
@@ -295,7 +295,9 @@ class Metaculus:
         if r.status_code < 400:
             return self.get_questions_for_pages(query_string, max_pages, current_page + 1, results + r.json()["results"])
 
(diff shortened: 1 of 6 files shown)


ergo/metaculus.py
```diff
@@ -188,8 +182,11 @@ def fit_single_logistic(self, samples):
             loc = min(max(loc, -0.1565), 1.1565)
             return loc, scale
 
-    @classmethod
-    def to_dataframe(self, questions: List["MetaculusQuestion"]):
+    def get_scored_predictions(self):
+        pass
```
stuhlmueller: Should this raise `NotImplementedError`?

stuhlmueller: Not crucial, but could add return types to some of these functions.

PR #37, opened 2020-03-31
Fit mixture of logistic distributions

stuhlmueller:
Next steps:

1. [x] Figure out a solution for when to stop fitting
2. [x] Use Adam optimizer instead of plain SGD - doesn't work as well, sticking with plain SGD
3. [ ] Use in `ContinuousQuestion.get_prediction`
4. [ ] Update Metaculus submission code to deal with multiple components
5. [x] Make Jax install work on Travis
6. [x] Add more tests for multiple components
7. [x] If necessary run with multiple random initializations, take best fit - decided not to for now, will revisit once I see cases where that's needed


brachbach:
@stuhlmueller For what it's worth I just created my own branch `br-mixture` based on this branch

when I'm done adding the ability to predict using mixtures, I'll make a PR to this branch so you can see what I added, and a PR to master to merge everything in

PR #58, opened 2020-04-15
Add foretold distributions to ergo

william-r-s:


william-r-s:
Looks like the CI failure is due to poetry, ignoring.

brachbach:
@william-r-s I'd think that the CI failure is somehow specific to this PR -- `poetry` and everything else have always worked consistently before

I think we should put some effort into fixing the CI failure before merging

ergo/foretold.py
```diff
@@ -0,0 +1,70 @@
...
+    """"Aggregated distribution from a foretold question"""
+
+    def __init__(self, id):
+        """
+            id: measurableId, the second id in the URL for a foretold question
+        """
+        self.id = id
+        self.floatCdf = None
+        # previousAggregate is the most recent aggregated distribution
+        response = requests.post(
```
brachbach, line 18: small thoughts:

1. this indentation is a bit weird/confusing to me, I'd indent more
2. I'd do the response/j bit as a method that gets called, to keep the `init` really clean simple
3. I'm not a fan of 1-character variable names like `j, I'd prob call it `json`, or go ahead and pull out the `data` or something

don't feel strongly, would actually be curious if you think either of those suggestions would in fact make the style worse

ergo/foretold.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def sample(self):
+        """Sample from CDF 
+
+        First sample between 0 and 1, find the corresponding bin, then linearly interpolate within the bin"""
+        xs = torch.tensor(self.floatCdf["xs"])
+        ys = torch.tensor(self.floatCdf["ys"])
+        y = uniform()
+        i = np.argmax(ys > y)
+        if i == len(ys) - 1:
```
brachbach, line 60: seems like when this condition is met, the x returned will be a bit lower than it should have been?

because in this case we're just taking the last xs[i], when ideally we'd want to add a bit to it, since ys > than the last y[i]

(not sure what to do about this, mostly just checking my understanding)

tests/test_foretold.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class TestForetold:
+    def test_foretold_sampling(self):
+        dist = ergo.ForetoldDistribution("cf86da3f-c257-4787-b526-3ef3cb670cb4")
+        num_samples = 20000
+        samples = ergo.run(
+            lambda: ergo.tag(dist.sample(), "sample"), num_samples=num_samples
+        )
+        assert np.count_nonzero(samples > 100) == pytest.approx(num_samples / 2, 0.1)
```
brachbach, line 13: maybe add a quick explanation of why we'd expect this to be true given how the Foretold distribution is set up?

and/or add a link to the foretold question?

ergo/foretold.py
```diff
@@ -0,0 +1,70 @@
...
+        return f"https://www.foretold.io/c/{self.channelId}/m/{self.id}"
+
+    def sample(self):
+        """Sample from CDF 
+
+        First sample between 0 and 1, find the corresponding bin, then linearly interpolate within the bin"""
+        xs = torch.tensor(self.floatCdf["xs"])
+        ys = torch.tensor(self.floatCdf["ys"])
+        y = uniform()
+        i = np.argmax(ys > y)
```
brachbach, line 59: nice, I would not have thought to get i in this way, pretty nifty!

ergo/foretold.py
```diff
@@ -0,0 +1,70 @@
...
+            return xs[i]
+        x0 = xs[i]
+        x1 = xs[i + 1]
+        y0 = ys[i]
+        y1 = ys[i + 1]
+        w = (y - y0) / (y1 - y0)
+        return x1 * w + x0 * (1 - w)
+
+    def plotCdf(self):
+        plt.plot(self.floatCdf["xs"], self.floatCdf["ys"])
```
brachbach, line 70: maybe we should default to using seaborn for everything for consistency? don't feel strongly

ergo/foretold.py
william-r-s: json is the name of a python module, so probably confusing to use as a variable name, but changed to something else

ergo/foretold.py
william-r-s: I actually had an off-by-one error in the whole thing. Now it makes more sense, because the first ys should be 0, so we never sample a lower value than 0.

tests/test_foretold.py
william-r-s: Done

ergo/foretold.py
william-r-s: Done

william-r-s:
Made requested changes. CI tests now passing (which I think was a fluke to begin with). Next thing is for @stuhlmueller to review.

PR #82, opened 2020-04-17
Format all the files using black

brachbach:
@stuhlmueller this is just running `black` over everything

IMO:

1. makes nothing worse
2. makes some things noticeably better
3. most of the changes I don't really care about

note that there are [a few](https://black.readthedocs.io/en/stable/installation_and_usage.html) things you can configure, if you really want to

I'm still in favor of switching to black and I wouldn't change any of the defaults

let me know what you think -- if you're into it then I can:

1. make sure that `flake8` is still the best choice for CI
2. change the README re: what formatter to use

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ install:
 
 script:
 - poetry run mypy .
+- poetry run flake8
 - poetry run black . --check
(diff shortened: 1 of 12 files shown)


stuhlmueller:
lgtm

PR #126, opened 2020-04-19
Add module-level docs for ergo.metaculus

brachbach:
to be able to do this in a nice way, also start reading env variables from a .env file

EVERYONE WILL NEED TO ADD THE .env FILE ONCE THIS IS MERGED


ergo/metaculus.py
```diff
@@ -1,3 +1,26 @@
...
+and submit predictions, via the API (https://www.metaculus.com/api2/)
+
+**Example**
+
+In this example, we submit our own prediction based on the community prediction.
+
+.. doctest::
+
+    >>> import ergo
+    >>> metaculus = ergo.Metaculus(username="oughtpublic", password="123456", api_domain="www")
```
stuhlmueller, line 12: We need to find a better solution than providing username and password in the code. Maybe read it from an environment variable and set up Travis to provide that environment variable.

brachbach:
@stuhlmueller I:
1. Removed username and password
2. expanded the explanation of the module-level example for metaculus.py
3. Created a notebook for examples from the documentation and linked to it from the index page of the documentation

Merge if you feel like it's ready

stuhlmueller:
Goal for [doctests](http://doc.pytest.org/en/latest/doctest.html) is to run, so a valid username/password is needed?

Would prefer not having the code from docs copied in a notebook. Could get out of sync, feels redundant.

.env-example
```diff
@@ -0,0 +1,3 @@
+METACULUS_USERNAME=oughttest
```
brachbach, line 1: The tests will only work if you use the oughttest account (because they use private questions from that account)

So, we want to mandate in the example .env file that you should use that account

Obviously this is a little weird, but I figured it was a pretty simple way to do this for now that's still very extensible

I think I'll go ahead and try to publish the test questions on Metaculus so we can do away with this soon -- @stuhlmueller , sound good?

PR #152, opened 2020-04-22
Change in community prediction over time

### LINKED ISSUE #73: Load/track/show sets of related questions

zjmiller:
`MetaculusQuestion` instances now have a method `add_change` that takes a timestamp as an argument and adds to the question's `data` dictionary the change in the question's community prediction since that timestamp.

This PR also includes a change to  the `MetaculusQuestion.to_dataframe` static method. The change allows the optional specification of extra columns to display in the dataframe (e.g., the "change" column that gets added to questions if you call `get_change`).

Closes #73


zjmiller:
I still need to make the change in community prediction more interpretable for continuous questions. But I'll plan to add this in a separate PR.

ergo/metaculus.py
```diff
@@ -140,28 +159,84 @@ def sample_community(self):
...
+        # one column, e.g., ergo.MetaculusQuestion.to_dataframe(qs, extra_columns="change")
+        # Because Python treats strings similarly to lists, this can lead to cryptic errors.
+        # So I'm adding this explicit error to prevent confusion.
+        if type(extra_columns) != list:
+            raise TypeError("extra_columns must be a list")
+
         show_names = any(q.name for q in questions)
         if show_names:
-            columns = ["id", "name", "title", "resolve_time"]
+            columns = ["id", *extra_columns, "name", "title", "resolve_time"]
```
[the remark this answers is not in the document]
zjmiller: Just so the columns that you're interested in aren't displayed on the far right of the DataFrame.

ergo/metaculus.py
```diff
@@ -125,6 +126,24 @@ def __str__(self):
...
+    def set_data(self, key: str, value: Any):
+        """
+        Add to data dict
+
+        :param key:
+        :param value:
+        """
+        self.data[key] = value
+
+    def add_change(self, since: int):
```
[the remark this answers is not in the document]
zjmiller: I changed it to `calculate_change` for now, because this implies it can be used not only to introduce a change value but to update it, but unlike `set_change` it doesn't imply that the caller of this method is determining the value.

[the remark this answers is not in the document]
zjmiller: If it were just for internal use, `set_change` might be preferable, but because this name will appear in public-facing models, I want to make sure it's easy to understand. Something like `question.calculate_change(since=yesterday)` just reads very idiomatically to me.

stuhlmueller:
Dan replaced pendulum with plain datetime a few PRs back (but didn't remove it from poetry dependencies). If you don't really need it, also use plain datetime and remove pendulum from dependencies?

zjmiller:
Sure I'll see if datetime can replace pendulum here.

ergo/metaculus.py
```diff
@@ -192,28 +201,81 @@ def sample_community(self):
         raise NotImplementedError("This should be implemented by a subclass")
 
     @staticmethod
-    def to_dataframe(questions: List["MetaculusQuestion"]) -> pd.DataFrame:
+    def to_dataframe(
+        questions: List["MetaculusQuestion"], extra_columns: List[str] = []
```
stuhlmueller, line 205: I think the API would be clearer if `to_dataframe` took an optional `columns` keyword argument that defaulted to an initial value like `["id", "name", "title", "resolve_time"]` but allowed users to overwrite it. This way, users can control all columns that appear and the order they appear in, and the calling code makes it obvious which columns will be part of the result.

ergo/metaculus.py
```diff
@@ -275,6 +337,20 @@ def score_prediction(self, prediction, resolution: float) -> ScoredPrediction:
             prediction["t"], prediction, resolution, score, self.__str__()
         )
 
+    def change_since(self, since: int):
```
stuhlmueller, line 340: Why is the `since` argument type `int` if it's a timestamp?

ergo/metaculus.py
```diff
@@ -192,28 +201,81 @@ def sample_community(self):
...
+                    question.id,
+                    *[question.data[key] for key in extra_columns],
+                    question.title,
+                    question.resolve_time,
+                ]
                 for question in questions
             ]
         return pd.DataFrame(data, columns=columns)
 
+    def get_i_of_community_prediction_before(self, timestamp: int) -> Optional[int]:
```
stuhlmueller, line 248: This function signature feels more natural to me:

```py
def get_community_prediction(self, before: datetime = None):
    ...
  

def change_since(self, since: datetime):
    """
    Calculate change in community prediction between the argument and most recent prediction
    :param since: timestamp
    :return: change in community prediction since timestamp
    """
    try:
        old = self.get_community_prediction(before=since, default=None)
    except LookupError:
        return 0
    new = self.get_community_prediction()
    return new - old
```

ergo/metaculus.py
```diff
@@ -192,28 +201,81 @@ def sample_community(self):
...
+            if timestamp_difference <= 0:
+                have_not_gone_far_back_enough = False
+            else:
+                j = j + 1
+
+        # The following happens if there aren't any community predictions yet.
+        if j == 0:
+            return None
+
+        return len(self.prediction_timeseries) - j
```
stuhlmueller, line 277: The logic here would be easier to follow if you factor out generic functionality using a function like [`find_lt`](https://docs.python.org/3/library/bisect.html#searching-sorted-lists).

ergo/metaculus.py
zjmiller: Will change to `float`.

stuhlmueller: Why not `datetime`?

zjmiller: I guess I was thinking a `timestamp` (which is a `float` in Python I think), partly because that's what the Metaculus API returns. I'm happy to use `datetime` instead.

ergo/metaculus.py
zjmiller: I ended up using `bisect.bisect_left` and things are a lot simpler.

zjmiller:
OK I'll fix the flake errors (docstrings too long).

PR #163, opened 2020-04-24
Metaculus questions default names

aehsani:
Simple first change. 
If a question doesn't have a name, it defaults to the question title.

### DIFF
--- a/ergo/metaculus.py
+++ b/ergo/metaculus.py
@@ -1118,6 +1118,8 @@ class Metaculus:
         :param name: a custom name for the question
         :return: A MetaculusQuestion from the appropriate subclass
         """
+        if not name:
+            name = data.get("title")
         if data["possibilities"]["type"] == "binary":
             return BinaryQuestion(data["id"], self, data, name)
         if data["possibilities"]["type"] == "continuous":


stuhlmueller:
Thanks @aehsani! Try merging master into your local branches and see if CI passes?

aehsani:
Were the tests failing because of the el-paso notebook?

stuhlmueller:
We recently changed our Travis environment to provide API keys as environment variables and it looks like your Travis runs aren't running in an environment that has those. Not sure what the best way is to address this, this could be a general problem with PRs that are made from forks and not local branches?

aehsani:
@stuhlmueller Looks like one way to do it would be to include the environment variables in the .travis.yaml file and encrypt them. Commit history looks like you guys did this, but decided against it. Looks like the easiest way, but if not, I can try setting up a separate Travis environment for my fork if you send me the API keys.

stuhlmueller:
@brachbach Why did we decide against including encrypted vars?

stuhlmueller:
Encrypted travis file actually wouldn't help [because](https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml):

> Encryption and decryption keys are tied to the repository. If you fork a project and add it to Travis CI, it will not have access to the encrypted variables.

brachbach:
@stuhlmueller to answer your question, I just couldn't get it to work quickly, and so I set up the variables through the UI instead. But yeah I think that's right re: it wouldn't help with PRs made from forks.

PR #166, opened 2020-04-24
Scrubbed a Credential

AABoyles:
Strongly advise [performing a password reset](https://www.metaculus.com/accounts/settings/).


stuhlmueller:
Thanks! (I also just did that on master and changed the password.)

PR #187, opened 2020-04-26
now no additional question refresh necessary

### LINKED ISSUE #125: Automatically call question.refresh_question() when necessary
I expected that this would work:

```py
questions = metaculus.get_questions(cat="internal--el-paso")
for question in questions:
    question.show_community_prediction()
```
But I needed this:

```py
questions = metaculus.get_questions(cat="internal--el-paso")
for question in questions:
    question.refresh_question()  # Load detailed question data
    question.show_community_prediction()
```

djthorne:
The Stupid Quick Fix to #125 (but does not address the more complicated requirements)

The query in get_questions_json ommits the following fields returned by the simpler query in get_question and refresh_question
simple query: `self.s.get(f"{self.api_url}/questions/{id}")`
additional fields: [author_name, prediction_histogram ,anon_prediction_count, last_read]

(Conversely the get_question_json query returns the field comment_count_snapshot, not present in the simple query (given very preliminary testing))

The present pull request just takes the union of the results but likely a more elegant solution can be found by investigating the queries and api responses more closely

### DIFF
--- a/ergo/metaculus.py
+++ b/ergo/metaculus.py
@@ -1272,6 +1272,11 @@ class Metaculus:
 
         questions = get_questions_for_pages(query_string, pages)
 
+        # add additional fields ommited from previous query
+        for i, _q in enumerate(questions):
+            _r = self.s.get(f"{self.api_url}/questions/{_q['id']}")
+            questions[i] = dict(_r.json(), **_q)
+
         if not include_discussion_questions:
             questions = [
                 q for q in questions if q["possibilities"]["type"] != "discussion"
--- a/ergo/metaculus.py
+++ b/ergo/metaculus.py
@@ -1273,9 +1273,9 @@ class Metaculus:
         questions = get_questions_for_pages(query_string, pages)
 
         # add additional fields ommited from previous query
-        for i, _q in enumerate(questions):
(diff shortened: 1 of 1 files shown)


ergo/metaculus.py
```diff
@@ -1272,6 +1272,11 @@ def get_questions_for_pages(
 
         questions = get_questions_for_pages(query_string, pages)
 
+        # add additional fields ommited from previous query
+        for i, _q in enumerate(questions):
```
stuhlmueller, line 1276: Why the underscore variable names?

djthorne: I'll just remove them if you don't like them stylistically. I am used to them, particularly for a single letter variable, as a kind of weak private/ indication that this is a transient variable. Why did I not use it for the `i`? Good question. Seems like I should at least be consistent, but somehow indexes of the resulting object are less transient than the _q which will be mutated and forgotten.

djthorne:
So this doesn't close #125, but is a quick fix to make using ergo more intuitive for now. The issue is asking for a clever lazy loading scheme. Just to check, have people experienced serious issues stemming from stupid upfront loading?

PR #188, opened 2020-04-26
updated plots to reflect name attribute change

djthorne:
Get wrapped titles back now that the question.name object is automatically set to be the title


ergo/metaculus.py
```diff
@@ -438,9 +438,7 @@ def _scale_x(self):
     @property
     def title_name(self):
```
stuhlmueller, line 439: Method title doesn't fit anymore

Issue #31, opened 2020-03-29
Switch to Numpyro/Jax


stuhlmueller:

stuhlmueller:
Numpyro issues this depends on:

- [Unobserved discrete latent variables](https://github.com/pyro-ppl/numpyro/issues/542)
- [Inference by enumeration](https://github.com/pyro-ppl/numpyro/issues/571)

Many variables in our forecasting models are discrete since they correspond to Bernoulli variables (resulting from binary propositions that people make forecasts on). The first of these issues would allow us to include those variables in our models. The second would allow us to compute exact predictions for models that only include combinations of such binary variables. Together these two issues would make Numpyro universal enough that we'd want to switch Ergo over to using Numpyro.

Du Phan says ETA for the Numpyro issues is a couple weeks.

Issue #154, opened 2020-04-22
Ergo tag doesn't support NaN values


william-r-s:
ergo.tag(float("nan"), "test") raises ValueError, because using the pytorch deterministic distribution mechanism doesn't support NaN values. It would still be useful to log NaN values.

stuhlmueller:
This is as special case of https://github.com/oughtinc/ergo/issues/144

Issue #144, opened 2020-04-21
Support distributions on dates and other rich values


stuhlmueller:
Currently fails because `ergo.tag` tries to convert to tensor

Issue #128, opened 2020-04-19
Better distributions and inference API


stuhlmueller:
For example:

- Inference: `infer(model) -> Distribution`
  - This replaces what's currently called `run`
  - Forward/importance-sample by default
  - Pass parameters for method:
    - `infer(model)`
    - `infer(model, method="mcmc")`
  - Model is a thunk
  - Return `EmpiricalDistribution` instance of `Distribution`
    - We can use this distribution in other models with `sample(empirical_dist)`
    - Note that this is a distribution on all named values, not on the return value
      - Could consider automatically (also) tagging return value
- Sampling from community distributions: `sample(question.crowd_dist)`
  - Perhaps support `sample(question)` as a shorthand
- More generally make distributions API similar to WebPPL
  - Initialize distributions with keyword arguments `sample(Normal(mu=0, sigma=1))`
  - Provide shorthand for samplers `normal(mu=0, sigma=1)` that creates distribution and samples from it
- Both `sample` and the sampler shorthands take an optional `name` keyword argument that is used to name the sampling site in the trace
- Wrap all [distributions](https://github.com/pyro-ppl/numpyro/tree/master/numpyro/distributions) from Numpyro

stuhlmueller:
Maybe do together with #31

stuhlmueller:
Consider making distributions `TorchDistribution` objects.

PR #194, opened 2020-04-27
Switch to Numpyro (WIP)

stuhlmueller:
This PR switches from using the probabilistic programming library Pyro to using Numpyro. This has the advantages that (1) we can use numpy operations and arrays everywhere and don't need to deal with torch tensors and (2) MCMC inference will be faster.

Closes https://github.com/oughtinc/ergo/issues/31


stuhlmueller:
1. [ ] Debug tests marked as xfail
2. [ ] Default to a global random seed, or otherwise figure out a user-friendly solution for random seeds
3. [ ] Find a more principled solution for `sample_normalized_community`
4. [ ] Add an example and test case with inference
5. [ ] Adapt/delete notebooks
6 . [ ] Profile, make sure it's not unnecessarily slow

stuhlmueller:
Mixture fitting failed even before switching to Numpyro, so I extracted the simpler test that demonstrates it [as a separate PR](https://github.com/oughtinc/ergo/pull/220).

stuhlmueller:
Next steps:
- [ ] Figure out why tests are stalling, fix
- [ ] Update notebooks
  - [ ] Figure out what to do about length 1 arrays vs floats
- [ ] Use `sample_shape` for getting many samples, not list comprehension
- [ ] Add inference test case

Maybe:
- [Better distributions and inference API](https://github.com/oughtinc/ergo/issues/128)
- [Support distributions on rich values](https://github.com/oughtinc/ergo/issues/144)
    - [NaN values](https://github.com/oughtinc/ergo/issues/154)
- [Stochastic structure](https://github.com/oughtinc/ergo/issues/9)
- Find a more principled solution for `sample_normalized_community`

Issue #189, opened 2020-04-27
Pull up-to-date data


brachbach:
For the data that we're pulling from @KrisMoore and @onlyasith, we're currently pulling from a static copy rather than their most up to date sheets.

We can easily switch to pulling their live data by using IMPORTRANGE in our Google Sheet that we're pulling from

This does run the risk that they'll change their data format and we'll be left high and dry, but we can ask them not to

PR #207, opened 2020-04-27
Pull live data from @onlyasith and @KrisMoore

brachbach:
Note that you can use the diff for `notebooks/el-paso/el-paso-no-output.ipynb` to actually see what changed in the notebook code!

But you should also load `notebooks/el-paso.ipynb` to see that:
1. Using the new data changes our predictions a bit (expected/good, nothing seems too weird)
2. Now many of the predictions hit `Encoutered nan gradient, stopping early` (not great. Seems to make a lot of the predictions flatter than they were before)

I'm not sure how to address 2, marking as a draft PR until we figure out the best way.


notebooks/el-paso/el-paso-no-output.ipynb
```diff
@@ -328,22 +328,6 @@
         "From [this](https://pandemic.metaculus.com/questions/4128/when-will-el-paso-county-texas-experience-its-first-peak-number-of-covid-infections/#comment-28304) comment. We made a copy of the model [here](https://docs.google.com/spreadsheets/d/119xLDClNoSIR0xe_svjXz44yjPj9wXyc_WZPiHHn3ac/edit#gid=1213113172)."
       ]
     },
-    {
```
brachbach, line 331: this is just deleting an extraneous cell that wasn't supposed to be there

notebooks/el-paso/el-paso-no-output.ipynb
```diff
@@ -500,7 +486,10 @@
         "            cases = projected_cases.loc[date, \"New cases\"]\n",
         "            if np.isnan(cases):\n",
         "                raise KeyError\n",
+        "            \n",
         "            # Add some (fairly arbitrary) uncertainty around this point estimate\n",
+        "            if cases == 0:\n",
```
brachbach, line 491: this had to be added because we now sometimes project 0 cases (we didn't before)

brachbach:
Fixes https://github.com/oughtinc/ergo/issues/189

PR #228, opened 2020-04-29
add ability to plot samples and distribution

### LINKED ISSUE #215: In show_prediction, have an option to show both samples form our model as well as samples from the logistic mixture we fit to the model results to submit our predictions
Right now, since you can only have one `samples`, you can only see one or the other.

Being able to see both would be helpful for debugging cases where the logistic mixture diverges from the model samples

### LINKED ISSUE #224: Compare multiple models (to each other and community predictions)
Expand show_prediction() such that it accepts a DataFrame of samples with each column corresponding to a model.

djthorne:
Fixes #224 
Fixes #215 

New signature of show_prediction() includes a parameter plot: str  which accepts ("samples", "fitted", "both") 

Additionally, if a dataframe of samples are passed it will now plot each column as a different model.


ergo/metaculus.py
```diff
@@ -698,67 +703,61 @@ def show_prediction(
...
-                    percent_kept=percent_kept,
-                    side_cut_from=side_cut_from,
-                )
-            )
 
-            df["prediction"] = self.denormalize_samples(df["prediction"])
-            df["community"] = self.denormalize_samples(df["community"])
+            if type(samples) == pd.DataFrame:
+                if plot != "samples" and samples.shape[1] > 1:
+                    raise ValueError(
```
brachbach, line 722: I'm a bit worried about how complicated this is...

ergo/metaculus.py
```diff
@@ -698,67 +703,61 @@ def show_prediction(
...
-            plot = self.density_plot(df, _xmin, _xmax, fill="#b3cde3", **kwargs) + labs(
-                x="Prediction", y="Density", title=self.plot_title
-            )
-            plot.draw()
+        df = pd.melt(df, var_name="sources", value_name="samples")  # type: ignore
+
+        plot = self.comparison_plot(df, xmin, xmax, **kwargs) + labs(
+            x="Prediction",
+            y="Density",
+            title=self.plot_title + "\n\nPrediction vs Community"
```
brachbach, line 756: I think this fixes an "API bug" where comparison plots sometimes had their titles left off

brachbach:
@djthorne, thanks for quickly whipping this together! So much functionality!

1. I agree with Olivier that it would be really nice to have testing for this code. I don't know of any better approach than just making some graphs and manually checking that they look right, like the now-outdated `TestVisualPandemic` tests. So I'd be happy with that. But definitely open to a better way.
2. I'm generally a bit worried about the complexity of the code. Olivier I think had some good ideas for how to clarify it a bit
3. This code seems to fix 2 bugs I ran into on master:
   a. plot title sometime not shown when comparing to community
   b. some other bug that I never figured out that meant that I couldn't plot most questions
FYI I'm going to start using this code now despite the issues mentioned to get around those bugs
4. bugs:
   a. There seems to be a bug if you try to show the fitted distribution for a date question, see question 4128 in the Colab
   b. It seems like everything is always graphed on a linear x scale, not log. I think generally Metacuus log questions should be graphed on a log scale, and this is particularly an issue for e.g. question 4153
5. We should fix the bugs of course, and also I think 1 and 2 are really important, as evidenced by the 3 bugs mentioned here

ergo/metaculus.py
```diff
@@ -698,67 +703,61 @@ def show_prediction(
         :param num_samples: number of samples from the community
         :param **kwargs: additional plotting parameters
         """
+        if plot not in ("samples", "fitted", "both"):
+            raise ValueError("side keyword must be either 'samples', 'fitted', 'both'")
```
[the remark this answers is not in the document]
djthorne: Changed the signature as you suggested, so this goes away 👍

ergo/metaculus.py
```diff
@@ -698,67 +703,61 @@ def show_prediction(
...
+                    raise ValueError(
+                        "For multiple predictions comparisons, only samples can be compared (The plot parameter can only be 'samples')"
+                    )
+                for col in samples:
+                    df[col] = self.normalize_samples(samples[col])
+            else:
+                df["samples"] = self.normalize_samples(samples)
 
-            df = pd.melt(df, var_name="sources", value_name="samples")  # type: ignore
+        if plot in ("fitted", "both"):
```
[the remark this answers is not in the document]
djthorne: This is a good idea! Incorporated into newest commit

ergo/metaculus.py
```diff
@@ -698,67 +703,61 @@ def show_prediction(
         :param num_samples: number of samples from the community
         :param **kwargs: additional plotting parameters
         """
+        if plot not in ("samples", "fitted", "both"):
```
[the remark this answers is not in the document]
djthorne: Your suggestion above removes this 👍

ergo/metaculus.py
djthorne: Yep, the change in question.name behavior broke a couple unexpected things (and I didn't know that the behavior was changed :) ). That is why it is important to have tests I reckon

djthorne:
@brachbach   
1. We definitely need better coverage of the plotting tests. I think that is a issue onto itself. My read of the current marching orders is we are more feature-focused so I think that will be part of a consolidation phase later. Happy to tackle it when it has the priority 
2. Yes, that is a good analysis. We've made decent strides on that front (my first pass at this was UGLY). At the moment we are able to use the same functions for all question data type. Now the only things that are tweaked at the question-data-class level are scale_x (to allow for log-scale) and custom plot code for date questions (no way around that I could find atm). But the complexity to testing ratio is such that I think it is far from robust. 

Let me figure out these bugs!

ergo/metaculus.py
djthorne: That makes sense. The least I could do is comment it a bit more. Let's talk about the basic idea though. It has to be something like this if we want to have the samples param be a flexible thing that takes in a 1-d sample or 2-d (dataframe) of samples. The alternative would be that we could make it less magic-y and make it explicit for the user with a separate keyword param  "models_df" or something. I kind of like the magic-y version, but I don't feel very strongly about it.

brachbach: Yeah I agree that the magic-y version seems better.

I think the main thing I was reacting to is that you can't do `plot_fitted` for multiple predictions. But I guess that's reasonable enough, makes sense to leave it as is at least for now

ergo/metaculus.py
brachbach: haha yeah :)

brachbach:
Cool, yeah, let's fix the bugs and then we can merge this!

Agree that testing should be a separate issue -- I created this one: https://github.com/oughtinc/ergo/issues/235

I think I also ran into another bug, which I started to poke at [here](https://colab.research.google.com/drive/1v5-5ATlsVMtFEO8O8vpVH1ILOz0RseQx#scrollTo=5Us0Tau4TLRY&line=1&uniqifier=1)

I poked at it just enough to think that it seems to be plotting-related, but I didn't look into it too deeply

Could you also fix that bug?

djthorne:
#236

djthorne:
Fixes #236

brachbach:
@djthorne looks like we still have [this bug](https://colab.research.google.com/drive/1v5-5ATlsVMtFEO8O8vpVH1ILOz0RseQx#scrollTo=5Us0Tau4TLRY&line=1&uniqifier=1) that I mentioned in my last comment, right?

brachbach:
@djthorne I'm gonna merge this, created a high-priority issue for the outstanding bug: https://github.com/oughtinc/ergo/issues/238

PR #231, opened 2020-04-30
El paso with latest ergo

brachbach:
Use latest ergo in the el-paso notebook. 1. Simplify associating Metaculus questions and models to predict on them to make this compatible with latest ergo. 2. Use plotting code from pr #228 to get around plotting bugs on master


brachbach:
I'm going to go ahead and merge this without review because:

1. The results look reasonable
2. it's late, and I think it's important for folks to start tomorrow building on this rather than the old notebook

brachbach:
(I also didn't wait for Travis to finish, but it's not even doing anything with the notebook code, anyway, so there's no point)

PR #242, opened 2020-05-02
Notebook cleanup

### LINKED ISSUE #241: Restructure Notebook-Specific custom package imports
We should be able to import notebook-specific custom packages without mucking about with file systems. My proposal is to put such utilities (currently in /notebooks/el-paso) in their own folder (atm I have it as /notebook_utilities') with the __init__ set up such that you can simply:

`from notebook_utilities import el-paso`
`import el-paso.texas_data`

or 
`from notebook_utilities.el-paso import *`

djthorne:
Fixes #241


djthorne:
Summary of conversation with Andreas "don't make it a separate project, make it a sub-project"

ergo/contrib/el_paso/brachbach.py
```diff
@@ -2,7 +2,7 @@
 from typing import Callable
 
 import pandas as pd
-import sklearn
+import sklearn  # type: ignore
```
stuhlmueller, line 5: Add it to `mypy.ini` instead

.travis.yml
```diff
@@ -5,7 +5,7 @@ python:
 before_install:
 - pip install poetry
 install:
-- poetry install
+- poetry install -E notebooks
```
stuhlmueller, line 8: Why does travis need the notebooks extras?

PR #243, opened 2020-05-02
created ipynbs dir with output-stripped nbs (and helper func)

### LINKED ISSUE #240: Agree on Collaboration Flow with Notebooks
Discussed the situation with @brachbach a bit. Here is my position and proposal (not a perfect solution, but the pragmatic choice at the moment):

# Recommendation:
we create a system in line with modern frameworks that 
1. have a 'src' folder (in our case perhaps 'notebooks' or 'ipynb') where we keep all of our notebooks **always saved without output**
2. We have a 'public' folder (in our case perhaps 'colabs' or 'notebooks' if we call the source 'ipynb') which has all the .ipynbs with outputs and with the colab link 
3. (Ideally) have a build system that creates 2. from 1. 

Note: we can move forward with this recommendation without 3. We can just do it manually for now (which is what we are de facto doing now anyway)

# Rational:
While seemingly close, the technology is not at a place where it makes sense to have the deliverable (nice colab with outputs pre-computed) the source files be the exact same file. Let's just accept that and it will make our lives easier. The good news is they are very close, so going back and forth isn't a big issue. This setup would allow for smooth local development (I think going back and forth is a moderate dev slower) and no outputs lets vanilla git flow work in many cases (no special tools needed). Additionally (with 3) we would have the assurance that WYSIWYG for the deliverables (they will have been just run from source so no gotcha's for consumers of the notebooks) 

# Alternative:
The alternative isn't significantly worse. We keep using the one directory as both base for continual development and the external deliverable. This just means that the dev flow is more difficult (strip out output on pull) and diffing with vanilla git is often not reasonable (which in some situations can be a decent headache). Additionally dev is responsible for ad-hoc making sure the ipynb renders into a collab appropriately. 

@stuhlmueller @brachbach Let me know your thoughts on this

djthorne:
fixes #240


djthorne:
Summary of conversation with Andreas "let's use build and src and add make scrub make command"

build/el-paso.ipynb
```diff
@@ -6953,7 +6953,7 @@
         "import warnings\n",
         "import functools\n",
         "from datetime import timedelta, date\n",
-        "from notebook_utilities.el_paso import *  "
+        "from ergo.conrtib.el_paso import *"
```
stuhlmueller, line 6956: Typo `contrib`

PR #246, opened 2020-05-03
No past peak

### LINKED ISSUE #205: Stop predicting that the peak of infections might possibly have occurred a while ago
I'm pretty sure that we know this is false, per #173, but we're still putting some probability mass on it.

Also look into the same issue for peak hospitalizations

brachbach:
Don't sample impossible values from the community prediction when predicting the peak of infections. Fixes #205.

(Note that the fitted distribution still puts some probability mass on the impossible outcomes, even though the model no longer does.)


stuhlmueller:
Functionality looks good. I refactored it to make it more readable. Review and merge if it looks good?

Issue #217, opened 2020-04-28
Try to make our predictions good enough that we need only sample from the community prediction 30% of the time or less (1st pass)


brachbach:
...but come by this honestly, i.e. always sample from the community distribution exactly as much as we think makes sense to make the best prediction.

Trying to drive the amount we mix in the community prediction as low as possible will force us to make our models better. And I think more use of explicit models is more in keeping with the goal of the El Paso challenge.

brachbach:
This is a first pass to get the relatively low-hanging fruit -- we'll probably want to do a 2nd pass later

brachbach:
Stop and share plan for improvement before making changes

brachbach:
Shared [the plan](https://colab.research.google.com/drive/17wgJIdX17-Rbm5VOVqQ78KF61BtHq_Og)

Will now proceeed with the following from the plan:

1. When we want to know the number of people in the hospital, in the ICU, or on a ventilator in the past, just look it up in historical data from @KrisMoore
  a. currently we're "predicting" the values for ICU, ventilation, etc. for past dates using the same methods as we do for future dates, even though @KrisMoore lists the information for past dates
2. Use historical El Paso data as the main source for predicting ICU patients and ventilated patients for future dates
  a. Currently, we're using historical ratios from other regions for hospitalized : ICU and ICU : ventilated to predict these in El Paso. Much of the historical data were gathered by Pablo; he advised us to switch to mostly using historical El Paso data, and that definitely seems to me like the right call
  b. [Slack thread about how to do this](https://ought-inc.slack.com/archives/C012Q4DS08H/p1588573559010300)
3. Change prediction of confirmed cases to predict with some non-zero probability that the peak hasn't already passed. Simplest way to do it would be to add more fuzziness to the @onlyasith predictions. (Currrently our model is 100% confident that the peak was on April 26)
  a. Switching to mostly using the Columbia model might solve this, which is why it isn't prioritized higher
4. Simple backtesting -- run the model from points in the past

brachbach:
Updated plan:

1. ~~DONE When we want to know the number of people in the hospital, in the ICU, or on a ventilator in the past, just look it up in historical data from @KrisMoore
  a. currently we're "predicting" the values for ICU, ventilation, etc. for past dates using the same methods as we do for future dates, even though @KrisMoore lists the information for past dates~~
2. When we know an answer from historical data, don't mix in the community prediction
  a. Mostly done
[...]

PR #259, opened 2020-05-05
Use historical ICU and ventilation data when we have it

brachbach:
also, create the "get_historical_data" function to provide one clear way to try to look up historical data


brachbach:
You can see how this pushes our ICU/ventilation-at-peak predictions towards the community predictions (the last 2 questions)

brachbach:
for question 4137 in the Colab run captured in the built version of the notebook, the gradient descent terminated early before fitting the samples well

notebooks/src/el-paso.ipynb
```diff
@@ -538,18 +558,44 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "@ergo.mem\n",
+    "def peak_hospitalized():\n",
```
stuhlmueller, line 562: I think it would be slightly cleaner if this function were just `peak_hospitalization_date` and in `max_30d_hospital_confirmed_for_peak` we look up the hospitalization count using `hospital_confirmed_for_date(peak_hospitalization_date())`

notebooks/src/el-paso.ipynb
```diff
@@ -272,6 +272,26 @@
...
+    "    \n",
+    "    Return the value or raise a KeyError if we don't have it.\n",
+    "    \"\"\"\n",
+    "    \n",
+    "    # prefer Texas government data over @KrisMoore compiled data\n",
+    "    try: \n",
+    "        value = el_paso_cases.loc[date, column_name]\n",
+    "        if np.isnan(value):\n",
+    "            raise KeyError\n",
+    "        return value\n",
```
stuhlmueller, line 288: A single `return value` at the end of the function would also work.

notebooks/src/el-paso.ipynb
```diff
@@ -558,27 +604,23 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "@question(4204)\n",
-    "def peak_icu_admissions():\n",
+    "@question(4204, community_weight=0.3)\n",
```
stuhlmueller, line 607: Ideally we'd only mix in the community distribution when we don't have historical data, right? Same for next question. (Fine not to do as part of this PR)

notebooks/src/el-paso.ipynb
brachbach: yeah that does seem better now that I think about it

notebooks/src/el-paso.ipynb
brachbach: 👍

notebooks/src/el-paso.ipynb
brachbach: yeah I think that's right. Definitely planning to revisit mixing in community prediction soon, but actually maybe it would make sense to not mix it in for historical data as part of this PR

notebooks/src/el-paso.ipynb
```diff
@@ -272,6 +272,26 @@
...
+    "    try: \n",
+    "        value = el_paso_cases.loc[date, column_name]\n",
+    "        if np.isnan(value):\n",
+    "            raise KeyError\n",
+    "        return value\n",
+    "    except KeyError:\n",
+    "        value = compiled_data.loc[date, column_name]\n",
+    "        if np.isnan(value):\n",
+    "            raise KeyError\n",
+    "        return value\n",
```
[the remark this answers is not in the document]
brachbach: good point :p

notebooks/src/el-paso.ipynb
```diff
@@ -587,19 +629,22 @@
...
-    "    return frac_icu_ventilation() * peak_icu_admissions()\n",
+    "    peak_hospitalized_date = peak_hospitalized()[0]\n",
+    "    \n",
+    "    try:\n",
+    "        return get_historical_data(peak_hospitalized_date, \"on_ventilator\")\n",
+    "    except KeyError:\n",
+    "        # If we don't have historical data,\n",
+    "        # use @Tamay's approach:\n",
+    "        # https://pandemic.metaculus.com/questions/4201/#comment-28004\n",
+    "        return frac_icu_ventilation() * peak_icu_patients()\n",
```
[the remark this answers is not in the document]
brachbach: I don't think I quite get what you mean.

The idea with `get_historical_data` is that you use it when you're hoping to be able to use historical data, but you're not sure whether we have it

and then if we don't, you'll have to make some prediction.

My impression was that using try-catch for this was pretty standard in Python, but I'm not sure -- I don't think I'd do it that way in JavaScript

notebooks/src/el-paso.ipynb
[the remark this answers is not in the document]
brachbach: Wait actually, this would always throw that exception on NaN, rather than failing over to the compiled_data on NaN if possible

But it's a good idea to have some sort of string with my custom KeyError, will implement that part of it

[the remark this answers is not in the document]
brachbach: (also BTW we're stuck on Python 3.6.9 because that's what's available in Colab)

notebooks/src/el-paso.ipynb
brachbach: (addressed here for these questions, added to https://github.com/oughtinc/ergo/issues/217 for other questions)

PR #260, opened 2020-05-05
Use Columbia model 

### LINKED ISSUE #190: Integrate Columbia model for confirmed cases (<3 hrs)
The Shaman lab at Columbia made [this model](https://github.com/shaman-lab/COVID-19Projection) for how COVID will play out at the county level in the US.

[Here](https://colab.research.google.com/drive/1fn72MFCnmz00y0NZTrNcYFBQ-a8EPAVe), BR and WS integrated the confirmed-cases predictions from the Columbia model

But, we haven't added this to the main notebook.

This would probably be a better replacement for some of the other models we're currently using, particularly @onlyasith's confirmed cases model (but also we should at least mix this model's predictions in for number of patients hospitalized, needing ventilation, etc.)

We should revisit how we've integrated this model to make sure we've done it in a sensible way, then merge it into master

djthorne:
Closes #190 

Use 'columbia model' as part of an ensemble to predict cases for a given day

Also refactor the model components section and add some descriptive text (please look at the text and see if it helpful/the right tone)


Makefile
```diff
@@ -15,9 +15,12 @@ format: FORCE  ## Run isort and black (rewriting files)
 docs: FORCE  ## Build docs
 	poetry run $(MAKE) -C docs html
 
-scrub: FORCE  ## Scrub notebooks/src/.ipynb of output
+scrub: FORCE  ## Create scrubbed files in notebooks/src/ from notebooks/build/ notebooks
 	poetry run python scripts/scrub_notebooks.py notebooks/build notebooks/src
 
+scrub_src: FORCE  ## Scrub notebooks/src/*.ipynb of output
```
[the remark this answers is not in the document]
djthorne: I'll describe it long form here and see if you have a better way to summarize it:
`scrub` takes what is in /build, scrubs it, and then overwrites the notebooks /src 
`scrub_src`  just scrubs the notebooks in /src

scripts/scrub_src.py
```diff
@@ -0,0 +1,45 @@
...
+import subprocess
+
+strip_metadata = {
+    "jupytext": {"notebook_metadata_filter": "-all", "cell_metadata_filter": "-all"}
+}
+
+strip_metadata_string = json.dumps(strip_metadata)
+
+
+def scrub(notebooks_path, scrubbed_path):
```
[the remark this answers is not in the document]
djthorne: I think this code is in flux for a bit until everyone uses the same workflow. Eventually there should be one file, but it is better to just give people options for now

notebooks/src/el-paso.ipynb
```diff
@@ -186,21 +190,47 @@
...
+    "    model = ergo.random_choice(models, weights)\n",
+    "    try:\n",
+    "        result = model(**params)\n",
+    "        if np.isnan(result):\n",
+    "            raise KeyError\n",
+    "        return result\n",
+    "    except:\n",
+    "        if fallback and len(models) > 1:\n",
+    "            i = models.index(model)\n",
+    "            del models[i]\n",
```
stuhlmueller, line 213: This mutates the arguments to `sample_from_ensemble`, could lead to unintended side-effects.

notebooks/src/el-paso.ipynb
```diff
@@ -186,21 +190,47 @@
...
+    "\n",
+    "    \"\"\"\n",
+    "\n",
+    "    model = ergo.random_choice(models, weights)\n",
+    "    try:\n",
+    "        result = model(**params)\n",
+    "        if np.isnan(result):\n",
+    "            raise KeyError\n",
+    "        return result\n",
+    "    except:\n",
```
stuhlmueller, line 210: Catch a more specific exception

brachbach:
I think you should move most of the Columbia model stuff into a Python file in `contrib/`, and then just import the functions that allow us to sample from different scenarios

notebooks/src/el-paso.ipynb
djthorne: My thinking is that we would want to catch all-cause model failure and then call the fallback models. Given that we won't know what errors the model may throw (outside the scope of the function) I have the omnibus catch. Happy to just catch the KeyError for now (we wrote the models and know that is enough) but then we should revisit this if this becomes a longer-lived function used across notebooks

notebooks/src/el-paso.ipynb
djthorne: @stuhlmueller So I should make copies and pass them in instead of mutating the original arguments?

notebooks/src/el-paso.ipynb
stuhlmueller: (The worry with catch-all is that there could be unexpected bugs that we don't notice.)

notebooks/src/el-paso.ipynb
stuhlmueller: Yes

PR #261, opened 2020-05-05
Install pendulum in prediction-dashboard colab

william-r-s:


stuhlmueller:
Also include the `build/` version with outputs since we're currently not automatically generating it, and instead generate the `src/` version by scrubbing the builds?

brachbach:
This didn't run, and also I realized that we needed to change the dates or else there'd be no output.

So closing this in favor of https://github.com/oughtinc/ergo/pull/292

Issue #226, opened 2020-04-29
Integrate all useful results from the Columbia model (<4 hrs)


brachbach:
In addition to confirmed cases, the Columbia model also estimates cumulative demand of hospital beds, ICU and ventilators.

We should factor those estimates into our own estimates of those parameters

brachbach:
Done in https://github.com/oughtinc/ergo/pull/267

PR #266, opened 2020-05-06
simplest 'select county' solution

### LINKED ISSUE #210: Generalize the El Paso model to all other counties covered by the Columbia data (< 8 hrs)
For the purpose of this issue, we should keep the "UX" of this really simple -- maybe, there's a cell in the notebook where you specify which country you want to run the model for (and the model runs for one county at a time)

TBC the thing I think we should do is keep it really simple, I'm not attached to that UX per se, it just seemed really simple

djthorne:
Closes #210 

I don't think this makes sense to merge without #226 (and other changes) as now all the other data is about el paso. Once we have the other variables based on the columbia model #226 it makes more sense, but we would have to factor out all the other data sources


brachbach:
@djthorne closing this because I think it's superseded by https://github.com/oughtinc/ergo/pull/278, LMK if that's wrong

Issue #204, opened 2020-04-27
Add more tests and checks to models


brachbach:
There's currently a lot of fairly complicated code in the Colab that isn't tested in any way. We should test it somehow.

This could look like graphing out all of the Model Components as sanity checks to be evaluated by a human, or it could be automated testing

brachbach:
@djthorne this seems like another one where general best practices from data science would be useful -- any thoughts?

brachbach:
@stuhlmueller based on my limited experience, I sort of feel like the problem of testing models well is a fairly core problem in data science -- it's just a lot harder to think of good tests than it is for, say, web server code.

Maybe this is a problem that Ergo should (eventually) try to provide an innovative solution to, or perhaps adopt someone else's innovative solution to

PR #275, opened 2020-05-11
small bug fixes in models

djthorne:
While working on generalization I would some potential bugs in the current question model section of the el-paso nb. @brachbach please make sure these make sense


brachbach:
@djthorne thanks for catching these!

I agree with Olivier's broad point that we'd really benefit from more testing of the models, see #204

On the bright side, I think that neither of these problems was actually coming into play, because I think it would have thrown an error if it were (but I didn't look into it too deeply)

PR #278, opened 2020-05-12
Apply El Paso model to multiple counties

### LINKED ISSUE #210: Generalize the El Paso model to all other counties covered by the Columbia data (< 8 hrs)

djthorne:
closes #210 
Additionally provides first rough pass at showing how a core model (the base county model) can be augmented with more data (in this case metaculus community predictions). It also refactors and extends /contrib/core with a way for interacting with non-metaculus questions. It is early days for that refactor and I expect we can consolidate it quite a bit. 

* covid-19.ipynb  
Allows you to choose any US county through a colab form. Then make predictions and graph them using @question decorator syntax consistent with what we had for metaculus.
currently it has county level data for cases and then get remaining 'primitives'---Hostpitalizations, ICU, and Ventilation from point estimate formulas using cases as an input. 

* covid-19-el-paso.ipynb
Displays the generalized county-based solution and shows how it can be augmented with the community distributions from metaculus with minimal changes. Can plot and compare the unaugmented and community augmented predictions. No additional data besides the community augmentation is currently utilized. 

* el-paso.ipynb: 
I expect currently does not work. I made breaking changes to contrib.  

I will need to fix el-paso before we merge (assuming we want to continue to support it)


notebooks/build/covid-19-el-paso.ipynb
```diff
@@ -0,0 +1,6756 @@
+{
```
brachbach: <p>I mentioned this elsewhere, but I think I'd like to fuzz the model projections rather than fuzzing how we use them</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/oughtinc/ergo/pull/278/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

brachbach: <p>I think we ideally want this and the "El Paso specific" distribution together on the same graph</p><p> </p><p>might work to have both of those plus the community distribution all on the same graph</p><p> </p><p>or if not, the community distribution could be on a separate graph</p><p> </p><p>actually what I think is most important is that they're vertically aligned and on the same scale, whether or not they're literally on the same graph doesn't actually matter that much</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/oughtinc/ergo/pull/278/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='44'/>

brachbach: <p>Is it just augmented with the Metaculus community prediction?</p><p> </p><p>Or does it also include the El Paso-specific historical data?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/oughtinc/ergo/pull/278/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='45'/>

notebooks/build/covid-19.ipynb
```diff
@@ -0,0 +1,4219 @@
+{
```
brachbach: <p>This looks pretty crazy -- would adding some fuzz to the Columbia predictions help?</p><p> </p><p>The other thing we could try to do is to reconstruct the full COlumbia distributions, since we have the percentiles -- that might be worth playing around with</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/oughtinc/ergo/pull/278/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='45'/>

brachbach: <p><code>max_30d_hospital_confirmed_for_peak</code>  looks really high, both just in general and compared to <code>peak_icu_patients</code>  -- any idea what's going on with that?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/oughtinc/ergo/pull/278/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='50'/>

notebooks/src/covid-19.ipynb
```diff
@@ -0,0 +1,798 @@
+{
```
brachbach: <p>Now that this is just for cases, the name and docstring of the function no longer make sense</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/oughtinc/ergo/pull/278/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

brachbach: <p>For the first sentence, I'd say, "In this notebook, we model some aspects of how the COVID-19 pandemic will play out in the US at the county level."</p><p> </p><p>(current version sounds like we're comparing different counties to each other, which we're not)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/oughtinc/ergo/pull/278/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

brachbach: <p>The docstring for <code>noisy_sma_infections</code>  shouldn't be the same as for <code>sma_infections</code></p><p> </p><p>but actually -- my guess is that we shouldn't have <code>noisy_sma_infections</code>.</p><p> </p><p>I'm guessing the reason to have it is that otherwise, there are only <code># of scenarios in the Columbia model * number of percentiles available from the model</code>  possible peaks? Or maybe fewer, perhaps as few as only 1?</p><p> </p><p>If so, then I think we should fuzz the model results, not <code>sma_infections</code></p><p> </p><p>(BTW I'm realizing that this notebook is a reasonably rough draft, which TBC you had already told me over Slack. So from here I'm going to stop commenting on things like the docstring and focus instead on the overall organization. I'll plan to do another pass later for things like docstrings once you've had time to clean up the notebook.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/oughtinc/ergo/pull/278/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

brachbach: <p>If possible, I'd like to import <code>myqs</code>  and then just use <code>@myqs</code>  as the decorator, not <code>@myqs</code></p><p> </p><p>Not totally sure what's feasible though and not a big deal</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/oughtinc/ergo/pull/278/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='43'/>

notebooks/build/covid-19-el-paso.ipynb
brachbach: I think the former, right?

brachbach:
@djthorne, thanks for getting all of this in, this is a meaty PR!

I think where we're currently at is:

1. We have somewhat-reasonable answers to the questions for counties beyond El Paso, in `covid-19.ipynb`.
  a. We could fairly easily make these answers somewhat better by either fuzzing the Columbia projections more or recovering the distribution of the Columbia projections rather than sampling from their percentiles
2. We have a notebook, `covid-19-el-paso.ipynb`, that uses the infrastructure that we have set up for all counties to figure out the answers to the questions for El Paso, then adjusts them based on the Metaculus predictions for El Paso. This is the only El Paso-specific data we're using; we're not using the historical data that we have for El Paso. So, we don't fully recover the predictions in `el-paso.ipynb`. That means we have a total of 3 notebooks re: COVID
3. The code from this PR is fairly rough, not yet at the level of `el-paso.ipynb`. (TBC, this is a reasonable result of @djthorne trying to get in a bunch of functionality in a short time)

I'd say ideally we'd want (decreasing order of confidence that it's worth it):

1. To fully recover the results from `el-paso.ipynb` in `covid-19-el-paso.ipynb`, such that we could get rid of `el-paso.ipynb`.
  a. We may want just one notebook, `covid-19-el-paso.ipynb`, with the El Paso specific stuff at the bottom. Or quite possibly 2 notebooks, one without the El Paso-specific stuff (`covid-19.ipynb`) and one with (`covid-19-el-paso.ipynb`), but it's very clear how the one with El Paso-specific stuff just picks up from the one without and builds on it (I think something like this was Dan's idea with `covid-19.ipynb` and `covid-19-el-paso.ipynb`)
2. All notebooks are at the level of presentability/niceness that `el-paso.ipynb` is at on master
3. More reasonable results for counties beyond El Paso
   a. don't add new data/modeling sources, but try to make better use of what we have, e.g. what I wrote above re: the Columbia projections

Practically, depending on how much time we want to spend, we may want to only do some of those things, or do them partway. Some specific ideas for that:
1. Basically stop where we are now (maybe a bit of cleanup, and definitely make sure that `el-paso.ipynb` still works). Our "flagship notebook" is `el-paso.ipynb`, but we say, "oh, by the way, we can also generalize to other counties", and we point to the other 2 notebooks
2. Focus on fully recovering the results from `el-paso.ipynb` in `covid-19-el-paso.ipynb` and then getting rid of `el-paso.ipynb`. Then we can say, "We have good predictions for El Paso, and you can see how we can also predict for other counties, though the predictions may not be very good." TBC I'd guess that this is more than half of the work to get to the "ideal state"

@djthorne what do you think of this, especially re: how I characterized what you've done so far and what we could still do?

@stuhlmueller, what do you think we should aim for?

djthorne:
@brachbach I think you characterized everything well. My expectation is that it would be a decent bit of work to elegently incorporate el-paso specific models/data when available (particularly if we want that notebook to also for for arbitrary counties). So that is why I was leaning towards the 3-notebook approach. I saw `covid-19-el-paso.ipynb` as showing 'here is how you take a naive model and augment it easily. In this case we use community predictions from Metaculus.'

brachbach:
My updated best guess for what we should do (after reading @djthorne's comment above and talking to him) is to have 2 nbs:

1. keep `el-paso.ipynb` basically the same as it is now, as our flagship El Paso notebook
2. combine `covid-19-el-paso.ipynb` into `covid-19.ipynb`. Then, `covid-19.ipynb`:
  a. allows you to predict for any county
     i. Smooth out the Columbia predictions (by fuzzing them or recovering the distributions and sampling from those) to make these predictions look more reasonable, but otherwise don't spend time optimizing them
  b. uses the special information we have about El Paso to make better predictions for El Paso. 
     ii. add in all of the El Paso-specific data to fully recover the predictions from `el-paso.ipynb`.

`covid-19.ipynb` will be rather long and ungainly, and we'll tolerate less clarity and polish than for `el-paso.ipynb`.

brachbach:
Seems unlikely at this point that we'll ever come back to the El Paso model and put in the effort required to merge this. So closing for now (can always reopen later if we do want to put in that effort)

PR #45, opened 2020-04-03
Submit mixture

brachbach:
- submit mixtures to Metaculus and visualize them
- relatedly, create `class LinearQuestion(MetaculusQuestion)` and `class LogQuestion(MetaculusQuestion)`
- also, unrelatedly -- moved metaculus tests into their own file
- Main outstanding issue: https://github.com/oughtinc/ergo/issues/53

[Visualizations of our predictions made using the code from this PR ](https://colab.research.google.com/drive/1r4FjkgmZt0Lu6t_0bPoVjmBQvXOQcTz3#scrollTo=LiwHiT3WezEu&line=1&uniqifier=1)

PR #309, opened 2020-05-19
Interval condition

brachbach:
- create IntervalCondition
- add tests of interval condition loss function

### DIFF
--- a/ergo/distributions/conditions.py
+++ b/ergo/distributions/conditions.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import List
+from typing import List, Optional
 
 from typing_extensions import TypedDict
 
@@ -33,10 +33,6 @@ class PercentileCondition(Condition):
         return f"There is a {round(self.percentile * 100)}% chance that the value is <{self.value}"
 
 
-HistogramEntry = TypedDict("HistogramEntry", {"x": float, "density": float})
-Histogram = List[HistogramEntry]
-
-
 @dataclass
 class HistogramCondition(Condition):
     histogram: Histogram
@@ -52,3 +48,20 @@ class HistogramCondition(Condition):
 
     def __str__(self):
         return f"The probability density function looks similar to the provided density function."
+
+
+@dataclass
+class IntervalCondition(Condition):
+    p: float
+    low: Optional[float]
+    high: Optional[float]
+    weight: float
+
+    def __init__(self, p, low=None, high=None, weight=1.0):
+        if low is not None and high is not None and high <= low:
+            raise ValueError(
+                f"High must be greater than low, got high: {high}, low: {low}"
+            )
+
+    def __str__(self):
+        return f"There is a {self.p:%.0} chance that "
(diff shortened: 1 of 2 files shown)


brachbach:
@stuhlmueller this is the absolute minimum I'd want to do in `ergo` to add IntervalCondition

the other thing to do would be to add more tests

Any thoughts on what tests I should add? (e.g. what tests for the other conditions I should duplicate but for IntervalCondition?)

brachbach:
We should consider either:
1. Replacing `PercentileCondition` with `IntervalCondition` with the low end open (should be pretty easy) or
2. not allowing open ends for `IntervalCondition` and converting an interval with one end open to a PercentileCondition

thoughts on which is better?

brachbach:
Implements #45

brachbach:
Implements https://github.com/oughtinc/elicit/issues/45

ergo/distributions/conditions.py
```diff
@@ -57,3 +60,31 @@ def loss(self, dist):
...
+
+
+@dataclass
+class IntervalCondition(Condition):
+    p: float
+    low: Optional[float]
+    high: Optional[float]
+    weight: float
+
+    def __init__(self, p, low=None, high=None, weight=1.0):
```
brachbach, line 72: the value for one end being open should maybe be +-infinity, rather than None

wasn't sure how likely infinity was to be annoying compared to None, I think both sort of make sense

brachbach: (switched to infinity, seems better I think)

tests/test_conditions.py
```diff
@@ -77,6 +83,32 @@ def test_weights():
     assert dist.components[0].loc == pytest.approx(2, rel=0.1)
 
 
+@dataclass
+class Uniform:
```
stuhlmueller, line 87: Would move to distributions folder, make instance of Distribution class. You can make an issue for that if you don't want to do it now.

ergo/distributions/conditions.py
```diff
@@ -57,3 +72,38 @@ def loss(self, dist):
...
+class IntervalCondition(Condition):
+    """
+    Condition that the specified interval should include
+    as close to the specified probability mass as possible
+
+    :raises ValueError: max must be strictly greater than min
+    """
+
+    p: float
+    min: float
```
stuhlmueller, line 87: I would reorder to `min, max, p` here, in `__init__`, and in elicit call. Defining bounds of interval first feels a bit more readable to me. Also wondering if it should be `xmin` and `xmax` but feel unsure.

tests/test_conditions.py
brachbach: Made this issue: https://github.com/oughtinc/ergo/issues/312

Was really just going for the minimum needed for testing here, we might want to add more to it for it to be a real distribution

ergo/distributions/conditions.py
brachbach: I put p first because it's required/has no default, since there is no obvious default

I'm not a huge fan of `xmin` `xmax`

I'd be into `lower_bound` `upper_bound`

I'm going to leave this as I had it for now, but we can change it later

PR #321, opened 2020-05-22
Normalize conditions before fitting

brachbach:
...then denormalize the resulting mixture

1. added lots of docstrings
2. added Condition.get_denormalized, even though we don't need it. I was already mostly done adding it when I realized we didn't need it, and it helps with the tests. Still might want to remove it later, at least on the abstract base class
3. add some tests. There could definitely be more

### DIFF
--- a/ergo/distributions/conditions.py
+++ b/ergo/distributions/conditions.py
@@ -186,14 +186,12 @@ class ScalePriorCondition(Condition):
         return self.weight * total_loss
 
     def get_normalized(self, true_min: float, true_max: float):
-        # I think this is wrong and that you do need to scale to normalize,
-        # but want to confirm with Andreas
-        return self
+        true_range = true_max - true_min
+        return self.__class__(self.weight, self.scale_mean / true_range)
 
     def get_denormalized(self, true_min, true_max):
-        # I think this is wrong and that you do need to scale to normalize,
-        # but want to confirm with Andreas
-        return self
+        true_range = true_max - true_min
+        return self.__class__(self.weight, self.scale_mean * true_range)
 
     def __str__(self):
         return f"The scale is normally distributed around {self.scale_mean}"
@@ -216,14 +214,12 @@ class LocationPriorCondition(Condition):
         return self.weight * total_loss
 
     def get_normalized(self, true_min: float, true_max: float):
-        # I think this is wrong and that you do need to scale to normalize,
-        # but want to confirm with Andreas
-        return self
+        true_range = true_max - true_min
+        return self.__class__(self.weight, (self.loc_mean - true_min) / true_range)
 
     def get_denormalized(self, true_min, true_max):
-        # I think this is wrong and that you do need to scale to normalize,
-        # but want to confirm with Andreas
-        return self
+        true_range = true_max - true_min
+        return self.__class__(self.weight, (self.loc_mean * true_range) + true_min)
 
     def __str__(self):
         return f"The location is normally distributed around {self.loc_mean}"
--- a/ergo/distributions/conditions.py
+++ b/ergo/distributions/conditions.py
@@ -27,7 +27,7 @@ class Condition(ABC):
         Return the condition normalized to [0,1]
 
         :param true_min: the true-scale minimum of the range
-        :param true_max: the true-scale minimum of the range
+        :param true_max: the true-scale maximum of the range
         """
 
     @abstractmethod
@@ -37,7 +37,7 @@ class Condition(ABC):
         Return the condition on the true scale of [true_min, true_max]
 
         :param true_min: the true-scale minimum of the range
-        :param true_max: the true-scale minimum of the range
+        :param true_max: the true-scale maximum of the range
         """
 
     def describe_fit(self, dist) -> Dict[str, Any]:
@@ -185,6 +185,16 @@ class ScalePriorCondition(Condition):
             total_loss += scale_penalty
         return self.weight * total_loss
 
+    def get_normalized(self, true_min: float, true_max: float):
+        # I think this is wrong and that you do need to scale to normalize,
+        # but want to confirm with Andreas
+        return self
+
+    def get_denormalized(self, true_min, true_max):
+        # I think this is wrong and that you do need to scale to normalize,
+        # but want to confirm with Andreas
+        return self
+
     def __str__(self):
         return f"The scale is normally distributed around {self.scale_mean}"
 
@@ -204,6 +214,16 @@ class LocationPriorCondition(Condition):
             loc_penalty = (self.loc_mean - component.loc) ** 2
             total_loss += loc_penalty
         return self.weight * total_loss
+
+    def get_normalized(self, true_min: float, true_max: float):
+        # I think this is wrong and that you do need to scale to normalize,
+        # but want to confirm with Andreas
+        return self
+
+    def get_denormalized(self, true_min, true_max):
+        # I think this is wrong and that you do need to scale to normalize,
+        # but want to confirm with Andreas
+        return self
 
     def __str__(self):
         return f"The location is normally distributed around {self.loc_mean}"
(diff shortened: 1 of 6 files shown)


ergo/distributions/conditions.py
```diff
@@ -123,6 +185,16 @@ def loss(self, dist):
             total_loss += scale_penalty
         return self.weight * total_loss
 
+    def get_normalized(self, true_min: float, true_max: float):
+        # I think this is wrong and that you do need to scale to normalize,
```
brachbach, line 189: @stuhlmueller is this comment accurate? I.e. do we need implement normalization for the 2 conditions you added?

(I think we should do it on a future PR if so, since they're not actually used in the code yet)

brachbach: Yep, it's definitely wrong, fixing now

ergo/distributions/conditions.py
```diff
@@ -20,6 +20,26 @@ def loss(self, dist):
         :param dist: A probability distribution
         """
 
+    @abstractmethod
+    def get_normalized(self, true_min: float, true_max: float):
+        """
+        Assume that the true range that the condition is over is [true_min, true_max].
+        Return the condition normalized to [0,1]
```
stuhlmueller, line 27: Could add this as `:return: return description` docstring

ergo/distributions/conditions.py
```diff
@@ -74,6 +94,26 @@ def describe_fit(self, dist):
         description["p_in_interval"] = float(self.actual_p(dist))
         return description
 
+    def get_normalized(self, true_min: float, true_max: float):
```
stuhlmueller, line 97: `scale_min`, `scale_max`, and `scale_range` would be more intuitive to me here. `self.min - true_min` was difficult to follow for me.

ergo/distributions/conditions.py
```diff
@@ -74,6 +94,26 @@ def describe_fit(self, dist):
...
+        true_range = true_max - true_min
+        normalized_min = (
+            ((self.min - true_min) / true_range) if self.min is not None else None
```
stuhlmueller, line 100: Could consider putting this in a function since it's used twice. Same in `get_denormalized`

ergo/distributions/location_scale_family.py
```diff
@@ -57,6 +57,19 @@ def sample(self):
         # FIXME (#296): This needs to be compatible with ergo sampling
         return self.odist.rvs(loc=self.loc, scale=self.scale)
 
+    def get_denormalized(self, true_min: float, true_max: float):
+        """
```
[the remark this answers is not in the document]
stuhlmueller: If you need to copy, can also consider a [copy_docs_from](https://github.com/pyro-ppl/numpyro/blob/f2aefcf1b034418244c1e465c64bce469388e566/numpyro/util.py#L238)-like decorator.

ergo/distributions/mixture.py
```diff
@@ -129,18 +139,39 @@ def from_conditions(
         initial_dist: Optional[M] = None,
         num_components: Optional[int] = None,
         verbose=False,
+        true_min=0,
+        true_max=1,
```
stuhlmueller, line 143: Kind of feels like this function has too many parameters now. Out of scope for this PR but we should refactor somehow.

ergo/distributions/mixture.py
```diff
@@ -236,3 +268,10 @@ def to_params(self):
             [c.loc, c.scale, weight] for c, weight in zip(self.components, self.probs)
         ]
         return np.array(list(itertools.chain.from_iterable(nested_params)))
+
+    def get_denormalized(self, true_min: float, true_max: float):
+        denormalized_components = [
+            component.get_denormalized(true_min, true_max)
+            for component in self.components
+        ]
+        return self.__class__(denormalized_components, self.probs, self.component_type)
```
stuhlmueller, line 277: Nice

ergo/platforms/metaculus/question/continuous.py
```diff
@@ -148,8 +148,20 @@ def community_dist(self) -> dist.HistogramDist:
         # TODO (#306): Unify distributions interface
         # TODO (#307): Account for values out of range in
         #   ContinuousQuestion.community_dist()
+        denormalized_max = float(
+            self.denormalize_samples(self.prediction_histogram[-1][0])
```
stuhlmueller, line 152: Could just call `true_from_normalized_value`, perhaps renamed to `denormalize_value`. Same in the two places below.

ergo/platforms/metaculus/question/continuous.py
```diff
@@ -148,8 +148,20 @@ def community_dist(self) -> dist.HistogramDist:
...
+        )
+        denormalized_min = float(
+            self.denormalize_samples(self.prediction_histogram[0][0])
+        )
+
+        denormalized_range_width = denormalized_max - denormalized_min
```
stuhlmueller, line 158: Why not just `denormalized_range`?

tests/test_conditions.py
```diff
@@ -42,6 +47,43 @@ def test_interval_condition():
     )
 
 
+def test_normalization_interval_condition():
+    condition = IntervalCondition(p=0.5, min=10, max=100)
+    assert condition.get_normalized(10, 1000).get_denormalized(10, 1000) == condition
```
stuhlmueller, line 52: Assign all 10s and 1000s in these functions to variables, then loop over a few different values for each?

ergo/distributions/conditions.py
brachbach: meh, the function would have too many params

ergo/distributions/location_scale_family.py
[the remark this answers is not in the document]
brachbach: I don't think I did copy paste the docstring in an inherited method -- which one were you thinking?

ergo/platforms/metaculus/question/continuous.py
brachbach: `true_from_normalized_value` seems to only exist on LogQuestion

(IIRC I made it to make it easier to implement `denormalize_samples` on LogQuestions)

ergo/distributions/conditions.py
stuhlmueller: What about something like this:

```py
@dataclass
class Scale:
    scale_min: float
    scale_max: float

    def range(self):
        return self.scale_max - self.scale_min
    
    def normalize(self, value):
        return (value - self.scale_min) / self.range() if value is not None else None
    
    def denormalize(self, value):
        ...


class FooCondition:

    def normalize(self, true_min: float, true_max: float):
        scale = Scale(true_min, true_max)
        normalized_min = scale.normalize(self.min)
        normalized_max = scale.normalize(self.max)
        return self.__class__(self.p, normalized_min, normalized_max, self.weight)
```

ergo/platforms/metaculus/question/continuous.py
stuhlmueller: What do you think about adding a `denormalize_value` to every Metaculus question class and using that?

ergo/distributions/conditions.py
brachbach: huh --- this sort of seems like a bunch more code and somewhat more abstraction for a slight gain in clarity, once you've read and understood it all

happy to defer to you on this, though, just let me know

stuhlmueller: I think across all of the conditions it will reduce code duplication enough to be worth it. Currently the line `true_range = true_max - true_min` appears 8 times in this file. (Not sure what I sketched is exactly the right solution, but I think there is a solution like this that is better than the current version.)

brachbach: ah got it, I previously misunderstood this somewhat (didn't understand that it was supposed to be able to apply to ~any condition)

I'll see if something like this can work

brachbach: FYI incorporating this into my `wrong-fitted-p` branch (needed to add normalization for mixtures, so figured I might as well clean everything up)

PR #324, opened 2020-05-23
Add normalization for logistic mixtures

brachbach:
2. test normalization/denormalization for logistic mixtures
3. create Scale and refactor all normalization to use it

### DIFF
--- a/ergo/distributions/conditions.py
+++ b/ergo/distributions/conditions.py
@@ -22,7 +22,7 @@ class Condition(ABC):
         """
 
     @abstractmethod
-    def get_normalized(self, scale_min: float, scale_max: float):
+    def normalize(self, scale_min: float, scale_max: float):
         """
         Assume that the condition's true range is [scale_min, scale_max].
         Return the normalized condition.
@@ -33,7 +33,7 @@ class Condition(ABC):
         """
 
     @abstractmethod
-    def get_denormalized(self, scale_min: float, scale_max: float):
+    def denormalize(self, scale_min: float, scale_max: float):
         """
         Assume that the condition has been normalized to be over [0,1].
         Return the condition on the true scale.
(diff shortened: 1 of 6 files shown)


ergo/distributions/location_scale_family.py
```diff
@@ -57,6 +58,20 @@ def sample(self):
         # FIXME (#296): This needs to be compatible with ergo sampling
         return self.odist.rvs(loc=self.loc, scale=self.scale)
 
+    def get_normalized(self, scale_min: float, scale_max: float):
```
stuhlmueller, line 61: Why `get_normalized` and `get_denormalized` instead of `normalize` and `denormalize`? The reason I can think of is to indicate that this doesn't mutate the existing instance and instead returns a new one, but I'd prefer to do this by convention for everything and clearly indicate the cases where we do mutate instances.

ergo/distributions/conditions.py
```diff
@@ -216,12 +209,12 @@ def loss(self, dist):
...
-        scale_range = scale_max - scale_min
-        return self.__class__(self.weight, (self.loc_mean - scale_min) / scale_range)
+        scale = Scale(scale_min, scale_max)
+        return self.__class__(self.weight, scale.normalize_point(self.loc_mean))
 
-    def get_denormalized(self, scale_min, scale_max):
-        scale_range = scale_max - scale_min
-        return self.__class__(self.weight, (self.loc_mean * scale_range) + scale_min)
+    def get_denormalized(self, scale_min: float, scale_max: float):
+        scale = Scale(scale_min, scale_max)
```
stuhlmueller, line 216: I wonder if the caller should pass a scale instead of a min/max.

ergo/distributions/location_scale_family.py
brachbach: yeah the non-mutating thing was exactly what I was thinking, but happy to go with non-mutating by default. Will rename.

ergo/distributions/conditions.py
brachbach: That'd be reasonable. I guess part of what I was thinking is that these functions (well maybe not these ones, but the LogisticMixture `normalize` function) might be called by application code (e.g. Elicit code), and it seems a bit weird to have to pass in a random Scale object in that case

brachbach: TBC I'm on the fence here and will change if you think it's better to change

stuhlmueller: Seems better to me to change, especially if it helps with log scales (not sure it does).

brachbach: Cool, I'll change it

PR #328, opened 2020-05-25
Add instructions to README for installation

steveryb:
I missed that the project used poetry, to adding here to help any other
contributors.

Also added a troubleshooting step to save the next developer ~3 hours of
python dependency hell. Turns out `poetry` will happily use an ancient
version of pip, any errors caused by that will show up as errors with your
python version, and there's no way by default to have poetry not do that.
The fix is to upgrade pip within the virtual environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,10 +38,20 @@ You can run the model [here](https://colab.research.google.com/github/oughtinc/e
 
 ## Quickstart
 
+### In your browswer
 1. Open [this Colab](https://colab.research.google.com/github/oughtinc/ergo/blob/master/notebooks/build/quickstart.ipynb)
 2. Add your Metaculus username and password
 3. Select "Runtime > Run all" in the menu
 4. Edit the code to load other questions, improve the model, etc., and rerun
+
+
+### On your computer
+1. Fetch and open the repo: `git clone git@github.com:oughtinc/ergo.git; cd ergo`
+2. If you don't have it installed already, [install Poetry](https://python-poetry.org/docs/#installation). Ergo uses this to manage its dependencies and environments.
+3. Use poetry to setup a virtual environment, and download dependencies: `poetry install`
+	- If you get `Could not find a version that satisfies the requirement jaxlib ...` after using poetry to install, this is probably because your virtual environment has old version of pip due to how [poetry choses pip versions](https://github.com/python-poetry/poetry/issues/732). Try `poetry run pip install -U pip`, then try `poetry install` again.
+4. Open a notebook: `poetry run jupyter notebook notebooks/build/quickstart.ipynb`
+5. Follow the instructions in the "In your browser" section to run the notebook.
 
 ## Philosophy
 


README.md
```diff
@@ -38,11 +38,21 @@ You can run the model [here](https://colab.research.google.com/github/oughtinc/e
...
 2. Add your Metaculus username and password
 3. Select "Runtime > Run all" in the menu
 4. Edit the code to load other questions, improve the model, etc., and rerun
 
+
+### On your computer
+1. Fetch and open the repo: `git clone git@github.com:oughtinc/ergo.git; cd ergo`
+2. If you don't have it installed already, [install Poetry](https://python-poetry.org/docs/#installation). Ergo uses this to manage its dependencies and environments.
+3. Use poetry to setup a virtual environment, and download dependencies: `poetry install`
+	- If you get `Could not find a version that satisfies the requirement jaxlib ...` after using poetry to install, this is probably because your virtual environment has old version of pip due to how [poetry choses pip versions](https://github.com/python-poetry/poetry/issues/732). Try `poetry run pip install -U pip`, then try `poetry install` again.
```
stuhlmueller, line 52: typo: "chooses"

stuhlmueller:
Looks helpful! Thanks Steven!

brachbach:
Thanks @steveryb! My high-level thoughts:

Right now, the main source for running `ergo` locally is [the docs](https://ergo.ought.org/en/latest/workflow.html) rather than the README. I think we should probably keep it that way. But:

1. We might want to link to them more prominently -- right now the only link within the README is at the very bottom
2. We might indeed want to include a quickstart in the README like this one

To help decide -- @steveryb, why did you want to get `ergo` running locally? Were you hoping to tweak it/do dev work on it? Or something else?

(asking to help map out a user journey from the README through to successfully and easily doing whatever you were trying to do, so that then we can put the most relevant documentation in the most helpful place)

README.md
```diff
@@ -38,11 +38,21 @@ You can run the model [here](https://colab.research.google.com/github/oughtinc/e
...
 1. Open [this Colab](https://colab.research.google.com/github/oughtinc/ergo/blob/master/notebooks/build/quickstart.ipynb)
 2. Add your Metaculus username and password
 3. Select "Runtime > Run all" in the menu
 4. Edit the code to load other questions, improve the model, etc., and rerun
 
+
+### On your computer
+1. Fetch and open the repo: `git clone git@github.com:oughtinc/ergo.git; cd ergo`
+2. If you don't have it installed already, [install Poetry](https://python-poetry.org/docs/#installation). Ergo uses this to manage its dependencies and environments.
+3. Use poetry to setup a virtual environment, and download dependencies: `poetry install`
```
brachbach, line 51: For what it's worth I think I've run into trouble at this step with `poetry` projects in the past because my Python environment [looked like this](https://xkcd.com/1987/) and I was accidentally trying to use some weird `miniconda` install.

So I'm not sure everyone will run into problems for the same reason, but still seems helpful to list out some common reasons for problems

brachbach:
@steveryb, just to be clear, I think the general state of Ergo is:

1. A few of us at Ought who have been working on it a lot are able to find our way around just because we've spent so much time with it
2. We've tried a bit to create documentation to help other devs use it, but we haven't tried very hard, and mostly we've been guessing at what would be helpful rather than relying on the actual experience of devs trying to use it
3. ...so now that you're learning to use it, it's a great opportunity for us to get feedback about how the README, docs, etc. do (not) help you get started, and how we can improve them

So please share any ideas you have!

steveryb:
Ah, so I completely missed that there were already docs for this. I think my standard pattern when trying to install something from github is to:
- scan the README for any information: I missed the link at the bottom, because I was expecting something like "install locally" or "getting started", rather than "Read more about Ergo development in the docs."
- look for something in the docs folder. Naively, I thought `workflow` meant more to do with what to do when you're already setup.

> To help decide -- @steveryb, why did you want to get ergo running locally? Were you hoping to tweak it/do dev work on it? Or something else?

I wanted to, in order:
1. help edit existing notebooks
2. start building my own notebooks using it
3. eventually start tweaking / doing dev work if it were useful

I *thought* I needed to have a local installation to easily get changes into a PR, looking through [the docs](https://ergo.ought.org/en/latest/workflow.html) it seems like that might be true, but I'm unsure.

Completely agree that having this in a single place, and linking to it more prominently would be great. I think we could just replace my text with a link to that doc and that would solve for my usecase. Happy to do that in this PR if you agree :)

brachbach:
Thanks @steveryb, that's helpful!

Seems like the user stories we should be sure to cover are:
1. I just want to make my own notebook using `ergo`. For this case, there's no reason to get `ergo` running locally, you can just start your own Jupyter notebook or Colab and `poetry install` ergo
2. I want to make a PR to change one of the notebooks in the `ergo` repo. I don't need to be able to change core `ergo` at all
3. I want to do dev work on core `ergo`

The stuff you need to know for each of those 3 is pretty different, and is sort of scattered. I think I'll take a shot at organizing it all and linking to it in the README, since I know where it is. Then I'll run it by you and see it makes sense to you

steveryb:
Sounds great! I think mentioning the `pip` version problem somewhere would be good,  but I'm assuming we can add that in after your refactor.

I'll abandon this PR then, feel free to assign me any future PRs on this.

steveryb:
(also happy to take any of those stories you want to unload)

brachbach:
awesome :)

I did end steal a good bit of your wording from this PR in the PR I just made: https://github.com/oughtinc/ergo/pull/330

PR #352, opened 2020-06-04
Normalize a histogram that does not cover the entirety of the true scale

brachbach:
NOTE: pausing work on this for now until either:
1. Dan's PR is merged and I can redo this on top of that, or
2. everything else about the [Elicit PR](https://github.com/oughtinc/elicit/pull/140) is ready to go, and we just need to wrap this up to merge it

Also, downscale the probability mass of a histogram for when it's being used to represent just part of a probability distribution

This is to support [this PR](https://github.com/oughtinc/elicit/pull/140)

This should have some tests, WIP


ergo/distributions/histogram.py
```diff
@@ -104,8 +104,54 @@ def sample(self):
...
-    def normalize(self):
-        return HistogramDist(self.logps, 0, 1)
+    def normalize(self, true_scale: scale.Scale = None):
+        """
+        Normalize the histogram onto [0,1]
+
+        Setting a true_scale allows you to express that the histogram to be normalized
+        did not cover the entire scale of interest.
+
+        E.g. -- imagine my histogram only has probability mass from 2 to 5,
```
brachbach, line 114: 1. I implemented this in `normalize` because that's the function I needed to call, and it's a fairly sensible place to put it. But arguably, this functionality should live on its own as `zero_pad` or something like that
2. The strategy here is to pad out the histogram with bins of zero probability, increasing the number of bins. Some possible alternatives:
a. keep the same number of bins, squeeze the probability into the middle, pad with zeros. I didn't want to do this because we'd have to destroy information (toss out some of the values we already have)
b. refactor HistogramDist so that it no longer assumes that the bins are evenly spaced. Then we'd only have to add a couple bins on either side of the dist to zero pad it
c. refactor HistogramDist and probably some other things (`CrossEntropyCondition`?) so that the HistogramDist expresses the fact that the histogram doesn't take up the entire true_scale in some other way than via its bins

@stuhlmueller, any thoughts on 1 or 2?

ergo/distributions/histogram.py
```diff
@@ -104,8 +104,54 @@ def sample(self):
...
+            true_scale.scale_min > self.scale_min
+            or true_scale.scale_max < self.scale_max
+        ):
+            raise ValueError(
+                "Can only rescale hist to a scale that includes all of its current scale"
+            )
+
+        x_range_below = scale.Scale(true_scale.scale_min, self.scale_min)
+        x_range_below_per_hist_range = x_range_below.range / self.scale.range
+        num_x_bins_below = round(self.size * x_range_below_per_hist_range)
```
brachbach, line 142: Note that this rounding introduces some imprecision. But it seems hard to avoid without one of the refactors mentioned above

ergo/distributions/histogram.py
```diff
@@ -194,14 +240,32 @@ def from_pairs(cls, pairs):
         logps = onp.log(onp.array(densities) / sum(densities))
         return cls(logps, scale_min=scale_min, scale_max=scale_max)
 
-    def to_pairs(self):
+    def to_pairs(self, total_p: float = 1.0):
```
brachbach, line 243: similarly to the changes to `normalize`, maybe this should live as its own `downscale_p` function outside of `to_pairs`

it could also just not even be part of `HistogramDist` and instead be something that's done to the pairs by Elicit

(I like that it's so incredibly easy to do this here, but would also be pretty easy to do in Elicit)

ergo/distributions/histogram.py
stuhlmueller: 2b. I think Dan is working on this as part of log scale PR.

https://github.com/oughtinc/ergo/compare/log_questions

ergo/distributions/histogram.py
```diff
@@ -104,8 +104,56 @@ def sample(self):
...
+        but I'm interested in p over [0,10].
+        So I'll use Scale(0,10), and I'll get back a histogram with additional
+        bins with 0 probability to cover the area from 0 to 2 and 5 to 10,
+        where my histogram doesn't place any probability.
+
+        :param true_scale: the full scale that I'm interested in probability over.
+        """
+        # if true_scale is not provided, assume that the histogram has
+        # entries over the entire scale of interest
+        if not true_scale:
```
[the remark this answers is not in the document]
brachbach: interesting idea --I don't think we'd want it to be a truthiness value for the scale, because the scale can be missing a `scale_min` or `scale_max` to indicate that that end of the scale is open/goes to +/-infinity

but maybe it could be a property like `scale.closed`

ergo/distributions/histogram.py
```diff
@@ -104,8 +104,56 @@ def sample(self):
...
+            return HistogramDist(self.logps, 0, 1)
+
+        if true_scale.scale_min is None or true_scale.scale_max is None:
+            raise ValueError(
+                "If you provide a true_scale, you must provide both a scale_min and a scale_max"
+            )
+
+        if (
+            true_scale.scale_min > self.scale_min
```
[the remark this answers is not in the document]
brachbach: Good idea!

ergo/distributions/histogram.py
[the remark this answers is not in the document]
brachbach: ohhhh got it

I was misinterpreting your comment, and actually my code is wrong/bad (basically I was accidentally applying a JS idiom to Python -- I know better but forgot as I was writing this)

thanks for clarifying

should be `if true_scale is None`

brachbach:
Closing because this was replaced by #371 .

Comments here might still plausibly be useful though

PR #357, opened 2020-06-05
Add TruncatedLogisticMixture

stuhlmueller:


ergo/distributions/logistic_mixture.py
```diff
@@ -3,151 +3,66 @@
...
             [c.loc, c.scale, weight] for c, weight in zip(self.components, self.probs)
         ]
-        return np.array(list(itertools.chain.from_iterable(nested_params)))
+        opt_params = np.array(list(itertools.chain.from_iterable(nested_params)))
+        return static.logistic_mixture_logpdf(opt_params, x)
 
     @classmethod
-    def from_params(cls, params):
-        structured_params = params.reshape((-1, 3))
+    def from_params(cls, fixed_params, opt_params, traceable=True):
```
brachbach, line 32: I'm confused by the fact that this method takes `fixed_params` and `traceable` but doesn't use them

maybe a Python idiom I'm not familiar with?

stuhlmueller: The reason is to keep the interface the same as in `Optimizable`. `HistogramDist` also implements `Optimizable` and does use those arguments.

PR #360, opened 2020-06-06
Add boxplot for COVID-19 tests

steveryb:
Forgot to commit the plot after running it again.


steveryb:
Not sure what our policy is for merging without review, so going to hold off until I get at least a cursory one.

PR #371, opened 2020-06-11
Add open boundaries [!pr]

brachbach:
A version of https://github.com/oughtinc/ergo/pull/352 that's built on more recent code and that we now use in the Elicit open-boundaries PR

### DIFF
--- a/ergo/distributions/histogram.py
+++ b/ergo/distributions/histogram.py
@@ -77,8 +77,59 @@ class HistogramDist(Distribution, Optimizable):
         num_bins = fixed_params.get("num_bins", 100)
         return onp.full(num_bins, -float(num_bins))
 
-    def normalize(self):
-        return HistogramDist(self.logps, scale=Scale(0, 1))
+    def normalize(self, true_scale: Scale = None):
+        """
+        Normalize the histogram onto [0,1]
+
+        Setting a true_scale allows you to express that the histogram to be normalized
+        did not cover the entire scale of interest.
+
+        E.g. -- imagine my histogram only has probability mass from 2 to 5,
+        but I'm interested in p over [0,10].
+        So I'll use Scale(0,10), and I'll get back a histogram with additional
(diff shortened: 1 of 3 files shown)


ergo/distributions/histogram.py
```diff
@@ -77,8 +77,59 @@ def initialize_optimizable_params(fixed_params):
         num_bins = fixed_params.get("num_bins", 100)
         return onp.full(num_bins, -float(num_bins))
 
-    def normalize(self):
-        return HistogramDist(self.logps, scale=Scale(0, 1))
+    def normalize(self, true_scale: Scale = None):
```
stuhlmueller, line 80: Feels like this should be split into two functions:

1. Expand the scale and add empty bins
2. Normalize

Is there a reason to merge these?

ergo/distributions/histogram.py
```diff
@@ -182,14 +233,29 @@ def to_lists(self, true_scale=True, verbose=False):
 
         return xs, ps
 
-    def to_pairs(
-        self, true_scale=True, verbose=False,
-    ):
+    def to_pairs(self, true_scale=True, verbose=False, total_p: float = 1.0):
```
stuhlmueller, line 236: Doesn't feel right to add this transformation here. Options that seem better:

1. Make a separate function that rescales the pairs
2. Support histograms that don't sum to 1 more generally, add a rescale method to HistogramDist that returns a rescaled histogram, then just export using the old `to_pairs` function

ergo/distributions/histogram.py
brachbach: nope!

ergo/distributions/histogram.py
brachbach: 1 seems good, I think 2 seems overengineered

brachbach:
@stuhlmueller I agree with your style comments, but if I don't end up having to touch Ergo again to improve Elicit, I may just leave this stuff as-is for now and do a follow-on PR to refactor

PR #374, opened 2020-06-15
Rewrite logistic mixtures so that components can be truncated

stuhlmueller:


ergo/distributions/logistic_mixture.py
```diff
@@ -1,67 +1,167 @@
...
-        # returns a normalized mixture naive to the true distribution
-        if scale is None:
+    def from_params(
+        cls, fixed_params, opt_params, scale=None, traceable=True
+    ):  # FIXME: traceable; why sometimes no Scale?
+        if not scale:
             scale = Scale(0, 1)
+        floor = fixed_params.get("floor", -np.inf)
+        ceiling = fixed_params.get("ceiling", np.inf)
+        # Allow logistic center to exceed the range by 25%
```
stuhlmueller, line 110: This should say 20%

PR #376, opened 2020-06-16
Fix Metaculus.get_questions so that it only excludes log date questio…

brachbach:
…ns, not regular log questions


brachbach:
This is a pretty obvious fix, and I used/tested it in my latest notebook, so I'm going to summarily merge once CI runs

PR #378, opened 2020-06-16
Move latest_community_percentiles 

brachbach:
from MetaculusQuestion to ContinuosuQuestion.

for ContinuousQuestions, this indeed gives the percentiles as desired. But for BinaryQuestions, the community_prediction is a float (rather than percentiles), so this returns a float

I could have added some replacement function to BinaryQuestion, but decided not to for now because it's unclear what we'd want. Binary question prediction_timeseries have the following structure:

```
{
    "t": 1592276813.6254,
    "distribution": {
        "avg": 0.81148,
        "num": 88,
        "var": 0.03514
    },
    "num_predictions": 124,
    "community_prediction": 0.87
}
```

We might want to return the `distribution` somehow, or maybe not, but seems unclear at the moment

### DIFF
--- a/ergo/platforms/metaculus/question/continuous.py
+++ b/ergo/platforms/metaculus/question/continuous.py
@@ -101,6 +101,15 @@ class ContinuousQuestion(MetaculusQuestion):
     def plot_title(self):
         return "\n".join(textwrap.wrap(self.name or self.data["title"], 60))  # type: ignore
 
+    @property
+    def latest_community_percentiles(self):
+        """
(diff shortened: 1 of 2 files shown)


brachbach:
I guess I think a regression is extremely unlikely in this case, because if you think about it it's pretty obvious that a BinaryQuestion doesn't have percentiles (I think someone just wasn't thinking about the distinction between MetaculusQuestion and BinaryQuestion when they added this initially). So not going to add a test

PR #383, opened 2020-06-18
Foundations for supporting open-bound questions

stuhlmueller:


ergo/conditions/crossentropy.py
```diff
@@ -25,3 +29,46 @@ def structure(cls, params):
 
     def __str__(self):
         return "Minimize the cross-entropy of the two distributions"
+
+
+class PartialCrossEntropyCondition(condition.Condition):
```
brachbach, line 34: I can't seem to find where this is actually used -- where is that?

stuhlmueller: Not used anymore - this was from my first attempt at open bounds.

PR #384, opened 2020-06-19
Log normalization fix [!pr]

uvafan:
Fixes log normalization such that the distance between 10 and 100 on log scale is the same as the distance between 100 and 1000. Screenshot:
image

Updated tests to reflect that `LogScale` would now only work for scales with a lower bound >0 (Metaculus requires questions with log scale have a lower bound >0).

Unsure why there's another test failing at the moment.

Automated elicit submission test results show little difference to those on master (except the ones that fail now which are mentioned below), which surprises me for this significant of a change. 
[Before](https://github.com/oughtinc/ergo/files/4802436/master_test.txt)
[After](https://github.com/oughtinc/ergo/files/4802439/after_test.txt)

Known issues: 
- Errors on [this question](https://www.metaculus.com/questions/2812/if-a-whole-human-brain-is-successfully-emulated-by-2100-what-is-the-maximum-number-of-humans-that-will-be-uploaded-in-any-5-year-period/) which has a max of 20B, which is larger than the max int Jax can hold (32-bit) so overflows.
- Somehow gets the error of submitting a loc less than the minimum (-2.047 < -2) on [this question](http://metaculus.com/questions/3892)


stuhlmueller:
I'm going to close this, assuming that @uvafan will integrate into #385

PR #385, opened 2020-06-22
Support Log Questions

stuhlmueller:
Elicit companion at https://github.com/oughtinc/elicit/pull/210


ergo/distributions/point_density.py
```diff
@@ -0,0 +1,199 @@
...
+            self.normed_densities = scale.normalize_densities(xs, densities)
+        if cumulative_normed_ps is not None:
+            self.cumulative_normed_ps = cumulative_normed_ps
+        else:
+            bin_probs = self.normed_bin_probs(
+                self.normed_xs, self.normed_densities, numpy=init_np
+            )
+            self.cumulative_normed_ps = init_np.append(
+                init_np.cumsum(bin_probs), init_np.array([1.0])
+            )
```
stuhlmueller, line 49: There should be a more elegant approach than appending `[1.0]`.

ergo/distributions/point_density.py
```diff
@@ -0,0 +1,199 @@
...
+    def from_conditions(cls, *args, fixed_params=None, scale=None, **kwargs):
+        if scale is None:
+            # TODO: Should we do this?
+            scale = Scale(0, 1)
+        if fixed_params is None:
+            fixed_params = {
+                "xs": np.linspace(
+                    scale.low, scale.high, constants.point_density_default_num_points
+                )
+            }
```
stuhlmueller, line 132: Should check where in the code base these can happen, address in the calling location, and consider requiring here.

ergo/distributions/point_density.py
```diff
@@ -0,0 +1,199 @@
...
+    @classmethod
+    def from_params(cls, fixed_params, opt_params, scale=None, traceable=True):
+        # TODO: traceable is always True here
+        if scale is None:
+            # TODO: Should we do this?
+            scale = Scale(0, 1)
+        xs = fixed_params["xs"]
+        ps = np.abs(opt_params)
+        Z = np.sum(cls.normed_bin_probs(xs, ps))
+        densities = ps / Z
```
stuhlmueller, line 146: It's unclear whether this is the right approach. We need some (fast) way to map real-valued parameters to a density function that integrates to 1.

ergo/distributions/point_density.py
```diff
@@ -0,0 +1,199 @@
...
+        return super(PointDensity, cls).from_conditions(
+            *args, fixed_params=fixed_params, scale=scale, **kwargs
+        )
+
+    @classmethod
+    def from_params(cls, fixed_params, opt_params, scale=None, traceable=True):
+        # TODO: traceable is always True here
+        if scale is None:
+            # TODO: Should we do this?
+            scale = Scale(0, 1)
```
stuhlmueller, line 142: As above, should check where in the code base this can happen, address it in the calling location, and consider requiring `scale` here.

ergo/distributions/point_density.py
```diff
@@ -0,0 +1,199 @@
...
+        )
+
+    @staticmethod
+    def initialize_optimizable_params(fixed_params):
+        num_points = fixed_params["xs"].size
+        return onp.full(num_points, 1.0)
+
+    @classmethod
+    def normalize_fixed_params(self, fixed_params, scale):
+        return {"xs": scale.normalize_points(fixed_params["xs"])}
```
stuhlmueller, line 159: `normalize_points` isn't vectorized, so this could lead to some pretty slow Jax compilations or slow compiled computations.

ergo/distributions/point_density.py
```diff
@@ -0,0 +1,199 @@
...
+        return -np.dot(self.normed_densities, self.normed_log_densities)
+
+    def cross_entropy(self, q_dist):
+        # We assume that the distributions are on the same scale!
+        return -np.dot(self.normed_densities, q_dist.normed_log_densities)
+
+    @classmethod
+    def normed_bin_probs(cls, normed_xs, normed_densities, numpy=np):
+        bin_sizes = numpy.diff(normed_xs)
+        bin_probs = (normed_densities[1:] + normed_densities[:-1]) / 2.0 * bin_sizes
```
stuhlmueller, line 198: We're taking the probability at the bin mean value here. Not sure this is compatible with what we're doing elsewhere.

ergo/scale.py
```diff
@@ -33,10 +33,10 @@ def denormalize_point(self, point):
         return (point * self.width) + self.low
 
     def denormalize_points(self, points):
-        return [self.denormalize_point(point) for point in points]
+        return np.array([self.denormalize_point(point) for point in points])
```
stuhlmueller, line 36: Should be vectorized. Same applies to `normalize_points`, `denormalize_densities`, and `normalize_densities`, both here and in `LogScale`.

tests/test_point_density.py
```diff
@@ -0,0 +1,124 @@
...
+        dist = PointDensity.from_pairs(pairs, scale)
+    elif dist_source == "structured":
+        dist = PointDensity.structure(direct_dist.destructure())
+    elif dist_source == "denormalized":
+        dist = direct_dist.normalize().denormalize(scale)
+    elif dist_source == "from_conditions":
+        condition = CrossEntropyCondition(p_dist=direct_dist)
+        dist = PointDensity.from_conditions(
+            [condition], fixed_params={"xs": xs}, scale=scale
+        )
```
stuhlmueller, line 50: The `from_condition` case does not work reliably.

tests/test_point_density.py
```diff
@@ -0,0 +1,124 @@
...
+        conditions = [IntervalCondition(p=0.5, max=value)]
+        dist = PointDensity.from_conditions(conditions)
+        assert dist.ppf(0.5) == pytest.approx(value, abs=0.1)
+
+
+@pytest.mark.parametrize("scale", scales_to_test)
+def test_density_pdf(scale: Scale):
+    uniform_dist = PointDensity.from_conditions([MaxEntropyCondition()], scale=scale)
+
+    assert uniform_dist.pdf(scale.denormalize_point(0.5)) != 0
```
stuhlmueller, line 91: All `!= value` checks should be replaced with something stricter.

ergo/distributions/point_density.py
```diff
@@ -0,0 +1,241 @@
...
+
+    @classmethod
+    def from_params(cls, fixed_params, opt_params, scale=None, traceable=True):
+        # TODO: traceable is always True here
+        if scale is None:
+            # TODO: Should we do this?
+            scale = Scale(0, 1)
+        xs = fixed_params["xs"]
+        # ps = np.abs(opt_params)
+        ps = nn.softmax(opt_params)
```
uvafan, line 170: Not sure if abs vs. softmax really matters here. I was experimenting with changing abs to softmax but it didn't make much of a difference as far as I could tell.

ergo/conditions/histogram.py
```diff
@@ -27,18 +27,22 @@ def loss(self, dist):
         return self.weight * total_loss / self.xs.size
 
     def normalize(self, scale: Scale):
-        normalized_xs = np.array([scale.normalize_point(x) for x in self.xs])
-        normalized_densities = np.array(
-            [density * scale.width for density in self.densities]
-        )
+        # TODO: Should we really just returns self here? Because the pdf and returned density are
```
djthorne, line 30: Functionally I think this is fine. Convention-wise, it also seems consistent. Is there anything in the way of dropping the scale argument in normalization in all conditions? Then it would be consistent with the convention in `distributions/` where normalize does not take a scale argument.

ergo/distributions/point_density.py
```diff
@@ -75,15 +80,24 @@ def pdf(self, x):
...
             dist = high_x - low_x
-            normed_density = np.where(dist == 0, low_density, (normed_x - low_x) / dist * high_density + (high_x - normed_x) / dist * low_density)
-            return normed_density / self.density_norm_term
+            normed_density = np.where(
+                dist == 0,
+                low_density,
+                (normed_x - low_x) / dist * high_density
+                + (high_x - normed_x) / dist * low_density,
+            )
+            return normed_density / self.scale.norm_term
```
uvafan, line 100: should we use Scale to denormalize this?

PR #388, opened 2020-06-26
Hist regression tests

brachbach:
Add regression tests for two bugs

The bugs are described in the test docstrings

These tests currently fail, so they're marked with xfail

We hope that the tests will pass once we switch to PointDensity

I think it makes sense to merge these now to give us some things to work towards for PointDensity


brachbach:
@stuhlmueller I'll merge this later today unless I hear from you

(very harmless, just adding some xfail tests)

PR #390, opened 2020-07-01
Generate consistent UTC timestamps

zjmiller:
Ergo currently uses the local timezone to turn a datetime (like the one obtained from "2020-06-30" via strptime) into a timestamp. This means it can assign different timestamps to the same Metaculus questions, depending on the timezone Ergo is operating in.

This PR changes Ergo to alway use UTC to generate the timestamp. 

This will partially fix the following Elicit issue:

Fixes OUG-209

### DIFF
--- a/ergo/platforms/metaculus/question/lineardate.py
+++ b/ergo/platforms/metaculus/question/lineardate.py
@@ -36,7 +36,10 @@ class LinearDateQuestion(ContinuousQuestion):
         return scale_x_datetime(limits=(xmin, xmax))
 
     def date_to_timestamp(self, date: str):
-        return datetime.datetime.strptime(date, "%Y-%m-%d").timestamp()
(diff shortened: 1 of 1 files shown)


stuhlmueller:
> Ergo currently uses the local timezone to turn a datetime (like the one obtained from "2020-06-30" via strptime) into a timestamp. This means it can assign different timestamps to the same Metaculus questions, depending on the timezone Ergo is operating in.

Is Metaculus also using UTC? @djthorne

djthorne:
@stuhlmueller The time strings we receive from Metaculus are without tz info (they are simply 'YYYY-MM-DD'). Given that I can only assume they are in UTC.

djthorne:
This is a good fix. Previously the timezone-unaware datetime object was converted to timestamp using local time (not UTC as it should for Metaculus). This fixes the issue!

PR #396, opened 2020-07-23
For fitting, use 400 points and no CDF interpolation

djthorne:
This PR improves a number of existing fitting bugs/quirks, but does not completely eradicate them:
[bug1](https://linear.app/ought/issue/OUG-270/point-density-has-blip-in-near-edge-intervals)
[bug2](https://linear.app/ought/issue/OUG-345/spike-fitting-issue-with-just-interval-maybe-caused-by-lowering-max)
[bug3](https://linear.app/ought/issue/OUG-263/interval-conditions-max-entropy-lead-to-spurious-intermediate-points)

It makes two updates:
1. Remove cdf interpolation
2. increased the grid density of Point Density distributions from 200->400

The performance hit for the denser grid has been negligible for typical elicit interactions (but clearly slower for logistic fitting)
The cdf interpolation will make things look much cleaner, but will decrease the actual accuracy slightly (this will be offset by the grid density increase)


ergo/distributions/point_density.py
```diff
@@ -46,16 +45,31 @@ def __init__(
...
+        return self.normed_densities * constants.bin_sizes
 
-        else:
-            self.cumulative_normed_ps = init_np.append(
-                init_np.array([0]), init_np.cumsum(self.bin_probs)
+    @cached_property
+    def normed_log_densities(self):
+        return np.log(self.normed_densities)
+
+    @cached_property
```
stuhlmueller, line 58: Why is this a cached property? Seems like it's doing its own caching?

pyproject.toml
```diff
@@ -49,6 +50,9 @@ nbdime = "^2.0.0"
 jupytext = "^1.4.2"
 ipywidgets = "^7.5.1"
 pytest-sugar = "^0.9.3"
+snakeviz = "^2.1.0"
+jupyter_http_over_ws = "^0.0.8"
+jupyterlab_github = "^2.0.0"
```
stuhlmueller, line 55: Are you intentionally adding this dev dependency?

ergo/distributions/point_density.py
```diff
@@ -217,39 +221,57 @@ def normalize_fixed_params(self, fixed_params, scale):
     # Export
 
     @classmethod
-    def add_endpoints(cls, xs, densities):
+    def add_endpoints(cls, xs, densities, scale=None):
```
stuhlmueller, line 224: (Out of scope) This could be a staticmethod.

ergo/distributions/point_density.py
```diff
@@ -217,39 +221,57 @@ def normalize_fixed_params(self, fixed_params, scale):
...
+        """
+        Exports the distribution in two arrays of xs and associated densities
+
+        :param denorm_xs_only: only denormalize the xs and leave the densities normalized
+        :param add_endpoints: add endpoints of the distribution to points passed out
+        :param num_points_out: the number of points to summarise the distribution with
+        :return: the true-scaled x values and densities (the normed density if denorm_xs_only=True)
+        """
+
+        if num_points_out:
```
stuhlmueller, line 260: Would use `if num_points_out is not None`

ergo/distributions/point_density.py
```diff
@@ -217,39 +221,57 @@ def normalize_fixed_params(self, fixed_params, scale):
...
+    def add_endpoints(cls, xs, densities, scale=None):
         """
-        Assumes xs are normalized. Returns a list of xs and densities
-        with endpoints that are on the edge of the scale.
+        Returns a list of xs and densities with endpoints that are on the edge of the scale
+        provided. If no scale is provided assume data is normalized.
         """
-        if xs[0] != 0:
+        if scale is None:
+            scale = Scale(0, 1)
```
stuhlmueller, line 230: Is this case used, or can we always require a `scale`?

ergo/distributions/point_density.py
```diff
@@ -217,39 +221,57 @@ def normalize_fixed_params(self, fixed_params, scale):
...
+
+        if xs[0] != scale.low:
             xdiff_ratio = (xs[1] - xs[0]) / xs[0]
             density = (densities[0] - densities[1]) / xdiff_ratio + densities[0]
             clamped_density = onp.maximum(density, 0)
```
stuhlmueller, line 235: (Out of scope) This logic is repeated below, would factor out

ergo/distributions/point_density.py
```diff
@@ -217,39 +221,57 @@ def normalize_fixed_params(self, fixed_params, scale):
...
-            xs = onp.append(xs, onp.array([1]))
+            xs = onp.append(xs, onp.array([scale.high]))
             densities = onp.append(densities, np.array([clamped_density]))
 
         return xs, densities
 
-    def to_arrays(self, denorm_xs_only=False, add_endpoints=False):
-        normed_xs = self.normed_xs
-        normed_densities = self.normed_densities
+    def to_arrays(self, denorm_xs_only=False, add_endpoints=False, num_points_out=None):
```
stuhlmueller, line 250: Would rename to `num_xs`

pyproject.toml
djthorne: I can remove them. What is the rational for what should be a dev dependency?

ergo/distributions/point_density.py
djthorne: are you referring to normed_log_densities? I'm sorry I don't follow

ergo/distributions/point_density.py
djthorne: we can always require scale

ergo/distributions/point_density.py
stuhlmueller: I meant `cumulative_normed_ps`

PR #398, opened 2020-07-28
Prevent errors due to computing log of probability 0

djthorne:
This PR has a simple fix for the situation where the density of a point/mass of a bin is 0. It adds a minuscule amount to the density prior when it is passed to a log function. 

An alternative would be to simply add this small amount to density proper, however I thought that there is end-user utility in being able to specify true 0 bins (trade-off being that there is a minuscule disconnect from the actual density and log density)

### DIFF
--- a/ergo/utils.py
+++ b/ergo/utils.py
@@ -76,5 +76,5 @@ def trapz(y, x=None, dx=1.0):
     return 0.5 * (dx * (y[..., 1:] + y[..., :-1])).sum(-1)
 
 
-def safe_log(x, constant=1e-15):
+def safe_log(x, constant=1e-37):
     return np.log(x + constant)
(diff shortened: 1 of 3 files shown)


ergo/utils.py
```diff
@@ -74,3 +74,7 @@ def trapz(y, x=None, dx=1.0):
     if x is not None:
         dx = np.diff(x)
     return 0.5 * (dx * (y[..., 1:] + y[..., :-1])).sum(-1)
+
+
+def safe_log(x, constant=1e-15):
```
stuhlmueller, line 79: Why did you choose 1e-15? (I would have expected a number closer to the minimum resolution)

```
>>> np.log(1e-37)
DeviceArray(-85.19565, dtype=float32)
>>> np.log(1e-38)
DeviceArray(-inf, dtype=float32)
```

Issue #147, opened 2020-04-22
Add PredictIt API


AABoyles:
It would be cool if ergo could expose an API for PredictIt markets so we could draw on PredictIt data and model it with (or against) Metaculus and Foretold,io questions.

stuhlmueller:
Good idea! Looks like they have [an API](https://predictit.freshdesk.com/support/solutions/articles/12000001878-does-predictit-make-market-data-available-via-an-api-) so should be pretty easy.

stuhlmueller:
A few prototypical use cases are:

1. I load one or more distributions from PredictIt, compute an implied distribution for a Metaculus question, and automatically submit the forecast to Metaculus.
2. I load one or more distributions from PredictIt, compute an implied distribution for another PredictIt question, perhaps transform it a bit (because I don't want to bet on my true beliefs, I want some profit margin), and then manually submit the distribution to PredictIt. Ideally this would also be automated, but PredictIt doesn't have a Write API so it's probably too much pain to support it.
3. I load one or more distributions from Metaculus, compute an implied distribution for a PredictIt question, and submit as in 2.

The next steps are:

1. Write out an example for each of 1-3 as code, pretending that we've already implemented the API.
2. Write a spec for the API based on what's needed to make these examples work well, with reasoning for why we made the choices we did.

PR #399, opened 2020-07-31
Implement PredictIt API

anthonyebiner:
### Purpose
Allows users to easily get market data from [the PredictIt API](https://www.predictit.org/api/marketdata/all). 
#147 

### Implementation
Adds three new classes: PredictIt, PredictItMarket, and PredictItQuestion. 

PredictIt is the main API wrapper class, allowing users to get markets from the PI API. Users can search for markets using either the id number (found in the url) or the name of the market. Unlike Metaculus, the PI API has a very strict ratelimit, making it unfeasible to make separate API calls for each market the user wants to get. Thus PredictIt makes a single call to get every market when it is instantiated. This market data can be refreshed with the given refresh method.

PredictItMarket is a class that represents a PI market. For example, the market [Who will win the 2020 Democratic presidential nomination?](https://www.predictit.org/markets/detail/3633/). A market contains several binary questions.

PredictItQuestion is a question in PI. Although a market title is a question, the various contracts in the market are the actual individual questions. For example, "Who will win the 2020 Democratic presidential nomination?" is a market, and "Biden" ("Will Biden win the 2020 Democratic presidential nomination?") is the question. I chose to structure it this way so that the question class most resembled a metaculus binary question class, however the name could be changed to PredictItContract to avoid confusion.

### Other Changes

- Added fuzzywuzzy package, allowing users to use shortened names to search for markets and contracts instead of id numbers. For example, a user can find the above market by searching "win 2020 dem". 
- Added classes to _init__.py files.

### Future work
Some PI questions are binned; they allow users to predict which bin a certain event will fall in. For example, the [popular vote margin market](https://www.predictit.org/markets/detail/6663/). Currently each individual bin is treated as an individual question. Future work could be done to make Ergo detect that it is, in fact, a binned market, and treat the whole market as one linear question.

### DIFF
--- a/ergo/platforms/predictit.py
+++ b/ergo/platforms/predictit.py
@@ -29,8 +29,7 @@ In this example, we grap the oldest PredictIt question and it's first contract a
     >>> print(name3 == name1)
     True
 """
-import re
-from typing import Any, Generator, List, Dict
+from typing import Any, Dict, Generator, List
 
 from dateutil.parser import parse
 import pandas as pd
@@ -46,7 +45,7 @@ class PredictItQuestion:
     :param market: PredictIt question instance
     :param data: Contract JSON retrieved from PredictIt API
 
-    :ivar PredictItQuestion question: predictit question instance
+    :ivar PredictItQuestion market: predictit market instance
     :ivar int id: id of the contract
     :ivar datetime.datetime dateEnd: end-date of a market, usually None
     :ivar str image: url of the image resource for the contract
--- a/ergo/platforms/predictit.py
+++ b/ergo/platforms/predictit.py
@@ -165,7 +165,7 @@ class PredictItMarket:
         :return: response
         """
         r = self.predictit.s.get(url)
-        if "Slow down!" in str(r.content):
+        if r.status_code == 429:
             raise requests.RequestException("Hit API rate limit")
         return r
 
@@ -236,7 +236,7 @@ class PredictIt:
         :return: response
         """
         r = self.s.get(url)
-        if "Slow down!" in str(r.content):
+        if r.status_code == 429:
             raise requests.RequestException("Hit API rate limit")
         return r
 
--- a/ergo/platforms/predictit.py
+++ b/ergo/platforms/predictit.py
@@ -1,35 +1,8 @@
 """
 This module lets you get question and prediction information from PredictIt
 and submit predictions, via the API (https://predictit.freshdesk.com/support/solutions/articles/12000001878)
-
-**Example**
-In this example, we grap the oldest PredictIt question and it's first contract and check to make sure the name's of both are consistent while we refresh.
-
-.. doctest::
-    >>> import ergo
-    >>> import os
-    >>> pi = ergo.PredictIt()
-
-    >>> market = list(pi.markets)[0]
-    >>> contract = list(market.questions)[0]
-
-    >>> name1 = market.name
-    >>> print(type(name1) is str)
-    True
-
-    >>> market.refresh()
-
-    >>> name2 = market.name
-    >>> print(name2 == name1)
-    True
-
-    >>> contract.refresh()
-
-    >>> name3 = contract.market.name
-    >>> print(name3 == name1)
-    True
 """
-from typing import Any, Dict, Generator, List
+from typing import Dict, Generator, List
 
 from dateutil.parser import parse
 import pandas as pd
@@ -42,7 +15,7 @@ class PredictItQuestion:
     """
     A single binary question in a PredictIt market.
 
-    :param market: PredictIt question instance
+    :param market: PredictIt market instance
     :param data: Contract JSON retrieved from PredictIt API
 
     :ivar PredictItQuestion market: predictit market instance
@@ -90,14 +63,6 @@ class PredictItQuestion:
             print(f"The column {name} could not be converted into a datetime")
             return dateEnd
 
-    def set_data(self, key: str, value: Any):
-        """
-        Set key on data dict
-        :param key:
-        :param value:
-        """
-        self._data[key] = value
-
     @staticmethod
     def to_dataframe(
         questions: List["PredictItQuestion"], columns=None,
@@ -113,12 +78,6 @@ class PredictItQuestion:
         data = [[question._data[key] for key in columns] for question in questions]
 
         return pd.DataFrame(data, columns=columns)
-
-    def get_community_prediction(self) -> float:
-        """
-        Return the lastTradePrice
-        """
-        return self.lastTradePrice
 
     def refresh(self):
         """
@@ -150,13 +109,14 @@ class PredictItMarket:
         :ivar str image: url of the image resource of the market
         :ivar str url: url of the market in PredictIt
         :ivar str status: status of the market. Closed markets aren't included in the API, so always "Open"
-        :ivar datetime.datetime timestamp: last time the market was updated.
+        :ivar datetime.datetime timeStamp: last time the market was updated.
             Api updates every minute, but timestamp can be earlier if it hasn't been traded in
     """
 
     def __init__(self, predictit: "PredictIt", data: Dict):
         self.predictit = predictit
-        self._load_attr(data)
+        self._data = data
+        self.api_url = f"{self.predictit.api_url}/markets/{self.id}/"
 
     def _get(self, url: str) -> requests.Response:
         """
@@ -172,23 +132,27 @@ class PredictItMarket:
     def __repr__(self):
         return f'<PredictItMarket title="{self.name}">'
 
-    def _load_attr(self, data):
+    def __getattr__(self, name: str):
         """
-        Load attributes of question from data to instance.
-        :param data:
+        If an attribute isn't directly on the class, check whether it's in the
+        raw contract data. If it's a time, format it appropriately.
+        :param name:
+        :return: attribute value
         """
-        self._data = data
-        self.id = data["id"]
-        self.name = data["name"]
-        self.shortName = data["shortName"]
-        self.image = data["image"]
-        self.url = data["url"]
-        self.status = data["status"]
-        self.api_url = f"{self.predictit.api_url}/markets/{self.id}/"
-        if data["timeStamp"] == "N/A":
-            self.timestamp = None
-        else:
-            self.timestamp = parse(data["timeStamp"])
+        if name not in self._data:
+            raise AttributeError(
+                f"Attribute {name} is neither directly on this class nor in the raw question data"
+            )
+        if name != "timeStamp":
+            return self._data[name]
+        dateEnd = self._data["timeStamp"]
+        if dateEnd == "N/A":
+            return None
+        try:
+            return parse(dateEnd)
+        except ValueError:
+            print(f"The column {name} could not be converted into a datetime")
+            return dateEnd
 
     @property
     def questions(self) -> Generator[PredictItQuestion, None, None]:
@@ -205,7 +169,7 @@ class PredictItMarket:
         used when the question data might have changed.
         """
         r = self._get(self.api_url)
-        self._load_attr(r.json())
+        self._data = r.json()
 
     def get_question(self, id: int) -> PredictItQuestion:
         """
--- a/ergo/platforms/predictit.py
+++ b/ergo/platforms/predictit.py
@@ -1,6 +1,6 @@
 """
 This module lets you get question and prediction information from PredictIt
-and submit predictions, via the API (https://predictit.freshdesk.com/support/solutions/articles/12000001878)
+via the API (https://predictit.freshdesk.com/support/solutions/articles/12000001878)
 """
 from typing import Dict, Generator, List
 
@@ -104,8 +104,8 @@ class PredictItMarket:
         :param predictit: PredictIt API instance
         :param data: Market JSON retrieved from PredictIt API
 
-        :ivar PredictIt predictit: predictit api instance
-        :ivar str api_url: url of the predictit api for the given question
+        :ivar PredictIt predictit: PredictIT API instance
+        :ivar str api_url: url of the PredictIt API for the given question
         :ivar int id: id of the market
         :ivar str name: name of the market
         :ivar str shortName: shortened name of the market
@@ -113,7 +113,7 @@ class PredictItMarket:
         :ivar str url: url of the market in PredictIt
         :ivar str status: status of the market. Closed markets aren't included in the API, so always "Open"
         :ivar datetime.datetime timeStamp: last time the market was updated.
-            Api updates every minute, but timestamp can be earlier if it hasn't been traded in
+            The API updates every minute, but timestamp can be earlier if it hasn't been traded in
     """
 
     def __init__(self, predictit: "PredictIt", data: Dict):
@@ -123,7 +123,7 @@ class PredictItMarket:
 
     def _get(self, url: str) -> requests.Response:
         """
-        Send a get request to to PredictIt API.
+        Send a get request to the PredictIt API.
 
         :param url:
         :return: response
@@ -164,7 +164,7 @@ class PredictItMarket:
         """
         Generate all of the questions in the market.
 
-        :return: iterator of questions in market
+        :return: generator of questions in market
         """
         for data in self._data["contracts"]:
             yield PredictItQuestion(self, data)
@@ -214,7 +214,7 @@ class PredictIt:
 
     def refresh_markets(self):
         """
-        Refetch all of the markets from the predictit api.
+        Refetch all of the markets from the PredictIt API.
         """
         self._data = self._get(f"{self.api_url}/all/").json()
 
@@ -223,7 +223,7 @@ class PredictIt:
         """
         Generate all of the markets currently in PredictIt.
 
-        :return: iterator of predictit markets
+        :return: iterator of PredictIt markets
         """
         for data in self._data["markets"]:
             yield PredictItMarket(self, data)
--- a/ergo/platforms/predictit.py
+++ b/ergo/platforms/predictit.py
@@ -2,18 +2,31 @@
 This module lets you get question and prediction information from PredictIt
 and submit predictions, via the API (https://predictit.freshdesk.com/support/solutions/articles/12000001878)
 
-pi = PredictIt()
+**Example**
+In this example, we predict the odds of democrats sweeping the house, senate, and presidency in the 2020 election based on the combined predictit odds.
 
-q_presparty = pi.search_market("party wins pres").search_question("dem")
-print(q_presparty.get_community_prediction())
-q_senparty = pi.search_market("who control senate").search_question("dem")
-print(q_senparty.get_community_prediction())
-q_houseparty = pi.search_market("who control house").search_question("dem")
-print(q_houseparty.get_community_prediction())
-q_demsweep = pi.search_market("dem sweep").get_question()
-print(q_demsweep.get_community_prediction())
-print("implied: " + str(q_senparty.get_community_prediction()*q_senparty.get_community_prediction()*q_houseparty.get_community_prediction()))
-print("given: " + str(q_demsweep.get_community_prediction()))
+We submit our prediction to the metaculus question asking "Will Democrats win both halves of Congress + Presidency in the US 2020 Election?"
+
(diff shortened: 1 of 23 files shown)


ergo/platforms/predictit.py
```diff
@@ -0,0 +1,332 @@
+"""
+This module lets you get question and prediction information from PredictIt
+and submit predictions, via the API (https://predictit.freshdesk.com/support/solutions/articles/12000001878)
+
+**Example**
+In this example, we predict the odds of democrats sweeping the house, senate, and presidency in the 2020 election based on the combined predictit odds.
```
stuhlmueller, line 6: This example is cool but I think it's going to break if either Metaculus or PredictIt questions change/close. Move it into a notebook under `notebooks/` and only add a doctest that we're pretty sure won't break in the next year or two?

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,332 @@
...
+    :ivar float lastTradePrice: last price the contract was traded at
+    :ivar float bestBuyYesCost: cost to buy a single Yes share
+    :ivar float bestBuyNoCost: cost to buy a single No share
+    :ivar float bestSellYesCost: cost to sell a single Yes share
+    :ivar float bestSellNoCost: cost to sell a single No share
+    :ivar float lastClosePrice: price the contract closed at the previous day
+    :ivar int displayOrder: position of the contract in PredictIt. Defaults to 0 if sorted by lastTradePrice
+    """
+
+    def __init__(self, market: Any, data: Dict):
```
stuhlmueller, line 65: Add "PredictItMarket" type for market. Ideally also a TypedDict type for data.

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,332 @@
...
+        self.market = market
+        self._data = data
+
+    def __repr__(self):
+        if self._data:
+            return f'<PredictItQuestion title="{self.name}">'
+        else:
+            return "<PredictItQuestion>"
```
stuhlmueller, line 73: When can this happen?

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,332 @@
...
+                except ValueError:
+                    print(
+                        f"The column {name} could not be converted into a datetime"
+                    )
+                    return self._data[name]
+            return self._data[name]
+        else:
+            raise AttributeError(
+                f"Attribute {name} is neither directly on this class nor in the raw question data"
+            )
```
stuhlmueller, line 100: If this is the same functionality as in MetaculusQuestion, should factor it out and reuse.

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,332 @@
...
+        """
+
+        data = [
+            [question._data[key] for key in columns]
+            for question in questions
+        ]
+
+        return pd.DataFrame(data, columns=columns)
+
+    def get_community_prediction(self):
```
stuhlmueller, line 129: Add return type (float?)

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,332 @@
...
+        :ivar str name: name of the market
+        :ivar str shortName: shortened name of the market
+        :ivar str image: url of the image resource of the market
+        :ivar str url: url of the market in PredictIt
+        :ivar str status: status of the market. Closed markets aren't included in the API, so always "Open"
+        :ivar datetime.datetime timestamp: last time the market was updated.
+            Api updates every minute, but timestamp can be earlier if it hasn't been traded in
+    """
+
+    def __init__(self, predictit: Any, data: Dict):
```
stuhlmueller, line 169: Add `"PredictIt"` type to `predictit`, `TypedDict` for `data`

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,332 @@
...
+
+    def refresh(self):
+        """
+        Refetch the market data from PredictIt,
+        used when the question data might have changed.
+        """
+        r = self.predictit._get(self.api_url)
+        self._load_attr(r.json())
+
+    def get_question_by_id(self, id: int) -> PredictItQuestion:
```
stuhlmueller, line 217: If there's not other method for getting questions, can just call it `get_question`.

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,332 @@
...
+        """
+        for data in self._data['contracts']:
+            yield PredictItQuestion(self, data)
+
+    def refresh(self):
+        """
+        Refetch the market data from PredictIt,
+        used when the question data might have changed.
+        """
+        r = self.predictit._get(self.api_url)
```
stuhlmueller, line 214: Shouldn't access a private property on `predictit`.

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,332 @@
...
+                return question
+        raise ValueError("Unable to find a question with that id.")
+
+    def search_question(self, name: str) -> PredictItQuestion:
+        """
+        Return the specified question given by the name of the question,
+        using fuzzy matching in the case where the name isn't exact.
+        :param name:
+        :return: question
+        """
```
stuhlmueller, line 234: This is cool, but I'm not sure it should be core functionality. Could move it to `contrib/predictit/fuzzy_search` and make fuzzywuzzy an [optional dependency](https://github.com/oughtinc/ergo/blob/master/pyproject.toml#L24).

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,332 @@
...
+        self._data = self._get(f"{self.api_url}/all/").json()
+
+    def _get(self, url):
+        """
+        Send a get request to to PredictIt API.
+        :param url:
+        :return: response
+        """
+        r = self.s.get(url)
+        assert "Slow down!" not in str(r.content), "Hit API rate limit"
```
stuhlmueller, line 271: Raise an Exception (either `RuntimeError` or something more specific) instead of failing assertion.

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,332 @@
...
+        return r
+
+    def refresh_markets(self):
+        """
+        Refetch all of the markets from the predictit api.
+        """
+        self._data = self._get(f"{self.api_url}/all/").json()
+
+    def get_markets(self) -> Iterator[PredictItMarket]:
```
stuhlmueller, line 280: Why `.get_markets()` here but `.questions` above? Pick one for consistency (probably property notation)?

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,332 @@
...
+        A market's id can be found in the url of the market.
+        :param id:
+        :return: market
+        """
+        for data in self._data['markets']:
+            if data['id'] == id:
+                return PredictItMarket(self, data)
+        raise ValueError("Unable to find a market with that ID.")
+
+    def search_market(self, name: str) -> PredictItMarket:
```
stuhlmueller, line 300: Would also factor this out to contrib.

stuhlmueller:
Two more thoughts:

- Add some unit tests
- Add documentation
- Check [these steps](https://ergo.ought.org/en/latest/core_contrib.html#before-submitting-a-pr)

ergo/platforms/predictit.py
anthonyebiner: It is slightly different, as it has to check to see if the value is 'N/A', and if it is, return a null date. It also uses "dateEnd" instead of "_time".

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,332 @@
...
+        if data['timeStamp'] == 'N/A':
+            self.timestamp = None
+        else:
+            self.timestamp = parse(data['timeStamp'])
+
+    @property
+    def questions(self) -> Iterator[PredictItQuestion]:
+        """
+        Generate all of the questions in the market.
+        :return: iterator of questions in market
```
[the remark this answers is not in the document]
anthonyebiner: Thanks for the review! Your comments were super helpful, especially the one about inverting the condition in __getattr__. The default implementation of __str__ just calling __repr__ slipped my mind too, so thanks for the reminder!

ergo/platforms/predictit.py
anthonyebiner: yep, they do give a 429. Changed to use the status code instead of checking for the slow down string.

stuhlmueller:
Thanks for the updates @anthonyebiner!

Next steps for Anthony:

1. Resolve the merge conflicts (by merging master into this branch)
2. Look over the pullrequest comments and address the important ones (but not critical since it only applies to contrib/)
3. Add a notebook with an example that shows how to use the API, including the functionality in contrib/

Then:

@brachbach reviews the PR, adds comments if he has more feedback, merges if it looks good.

stuhlmueller:
From the notebook:

> Interestingly enough, this is much lower than the odds of a sweep in the specific clean sweep market.

Could note that this is almost certainly because the outcomes for the individual markets are correlated - e.g. conditional on democrats winning house and senate, they're more likely to also win the presidency. Multiplying the individual market probabilities assumes that they're uncorrelated.

tests/test_predictit.py
```diff
@@ -0,0 +1,93 @@
...
+        "url": str,
+        "status": str,
+        "timestamp": datetime.datetime,
+    }
+    for market in predictit_markets:
+        for attr in attrs.items():
+            assert type(getattr(market, attr[0])) is attr[1]
+
+
+def test_market_questions(predictit_markets):
```
brachbach, line 38: I haven't looked carefully at this test and the tests below, but I'd guess that if I were to look at them more carefully, I'd feel confused about what they were testing.

I think to improve the tests, you should:
1. Delete some of them (if you want)
2. For each test, identify: what crucial functionality are we relying on the Predict API having? I think plausibly the tests that you'd want will look pretty different from what you have, such that you'll want to basically start over with writing the tests.
3. Explain what that functionality is on a docstring on the test
4. Make sure that the test directly tests that crucial functionality

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,268 @@
...
+        :param columns: list of column names as strings
+        :return: pandas dataframe summarizing the questions
+        """
+        if columns is None:
+            columns = ["id", "name", "dateEnd"]
+        data = [[question._data[key] for key in columns] for question in questions]
+
+        return pd.DataFrame(data, columns=columns)
+
+    def get_community_prediction(self) -> float:
```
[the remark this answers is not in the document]
brachbach: @stuhlmueller what do you think?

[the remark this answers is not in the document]
stuhlmueller: Yeah, probably shouldn't try to standardize for now. I think it would be nice to have a uniform interface eventually but the current middle ground between standardized and custom isn't great. Better to start with platform-specific implementations and then consider whether and how to standardize later.

tests/test_predictit.py
```diff
@@ -0,0 +1,93 @@
+import datetime
+
+import pandas as pd
+
+import ergo
+
+
+def test_refresh(predictit):
```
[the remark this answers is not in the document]
anthonyebiner: The problem with trying to test refreshing is that calling refresh doesn't actually guarantee any changes because the api is only updated once every minute. My reasoning behind the test was to make sure the program doesn't fetch a different market or question upon calling refresh by messing with the id.

anthonyebiner:
Thanks for the comments @brachbach! I'll be sure to implement those fixes and changes. One of the problems I'm running into is designing the test suite, though. The biggest stumbling block is that there is no guarantees for the specific data the API returns. The markets are always changing, and although the overall structure of the data remains the same, the actual content doesn't. This makes it difficult to write real-world tests that compare results to the website itself. 

Because of this, my mindset behind the test suite was to make sure the data the wrapper spits out has the same general stucture as the API, hence the tests that check things like attribute types. Other tests, such as test_markets, compare the data returned by two different methods that can be used to get the same thing, like using the .markets attribute or the get_market method.

Besides just testing to make sure the methods and attributes are returning an object of the correct type, I'm not sure how to go about writing tests that directly test crucial functionality. Being able to load a PredictIt object with a cached version of the API could solve some problems, but I'm not sure how to go about doing that without messing with the PredictIt classes. 

Any help would be greatly appreciated! I'm not too experienced with writing test suites...

tests/test_predictit.py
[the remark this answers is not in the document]
brachbach: cool -- I'd say:
1. Is there anything that *is* guaranteed to change on refresh? If so, test that
2. If not, just delete this test

tests/test_predictit.py
```diff
@@ -0,0 +1,93 @@
...
+    predictit.refresh_markets()
+    assert "markets" in predictit._data
+
+
+def test_markets(predictit):
+    markets = list(predictit.markets)[0:10]
```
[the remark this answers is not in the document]
brachbach: I can't think of a way to effectively test `markets`, I think just delete this test

tests/test_predictit.py
```diff
@@ -0,0 +1,93 @@
...
+    for market in markets:
+        assert type(market) is ergo.PredictItMarket
+        market_other = predictit.get_market(market.id)
+        assert market.name == market_other.name
+
+
+def test_market_attributes(predictit_markets):
+    attrs = {
```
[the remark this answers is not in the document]
brachbach: I think this test is prob fine as is, but add a docstring to explain what it's testing

tests/test_predictit.py
brachbach: I'd just delete this test

tests/test_predictit.py
```diff
@@ -0,0 +1,93 @@
...
+    for market in predictit_markets:
+        questions = list(market.questions)
+        assert len(questions) >= 1
+        for question in questions:
+            assert type(question) is ergo.PredictItQuestion
+            question_other = market.get_question(question.id)
+            assert question.name == question_other.name
+
+
+def test_market_refresh(predictit_markets):
```
brachbach, line 48: delete this one

tests/test_predictit.py
```diff
@@ -0,0 +1,93 @@
...
+    market = predictit_markets[0]
+    name = market.name
+    market.refresh()
+    assert market.name == name
+
+
+def test_question_refresh(predictit_markets):
```
brachbach, line 55: delete this one

tests/test_predictit.py
```diff
@@ -0,0 +1,93 @@
...
+    market = predictit_markets[-1]
+    question = list(market.questions)[0]
+    name = question.name
+    question.refresh()
+    assert question.name == name
+
+
+def test_question_dataframe(predictit_markets):
```
brachbach, line 63: for this one, you could test that dataframe has the expected columns

but also fine to just delete

tests/test_predictit.py
```diff
@@ -0,0 +1,93 @@
...
+    for market in predictit_markets:
+        questions = list(market.questions)
+        df = ergo.PredictItQuestion.to_dataframe(questions)
+        assert type(df) is pd.DataFrame
+
+
+def test_question_community_prediction(predictit_markets):
```
brachbach, line 70: I'd just delete this one, but if you like you can add a docstring explaining what it's testing

and maybe also test the community prediction is between 0 and 1?

tests/test_predictit.py
```diff
@@ -0,0 +1,93 @@
...
+    for market in predictit_markets:
+        for question in market.questions:
+            assert type(question.get_community_prediction()) is float
+
+
+def test_question_attributes(predictit_markets):
```
brachbach, line 76: for this one, I'd add a docstring explaining what you're testing

also, I don't think it's necessary to test every question in every market, just one or 2 would be fine

(but fine to test them all if the test runs quickly)

tests/conftest.py
```diff
@@ -175,6 +175,16 @@ def point_densities():
...
 
 
+@pytest.fixture(scope="module")
+def predictit():
+    return ergo.PredictIt()
+
+
+@pytest.fixture(scope="module")
+def predictit_markets():
+    return list(ergo.PredictIt().markets)[0:10]
```
brachbach, line 185: I don't think you really need 10, I think just 3 would be sufficient

notebooks/predictit_clean_sweep.ipynb
```diff
@@ -0,0 +1,380 @@
...
+        "colab_type": "text"
+      },
+      "source": [
+        "In this example, we use the PredictIt odds of democrats winning the house, senate, and presidency to make a prediction on whether they will sweep all three.\n",
+        "\n",
+        "PredictIt is composed of markets, which in turn are composed of questions. Although a PredictIt market has a literal question as its name, for consistency with Metaculus and Foretold, ergo considers each option in a market to be a separate \"question\". \n",
+        "\n",
+        "For instance, in the PredictIt market \"Which party will win the 2020 U.S. presidential election?\", one such question would be the democrat option. This is a binary question that essentially asks \"will democrats win the 2020 U.S. presidential election\".\n",
+        "\n",
+        "One thing to keep in mind is that the 10% profit fee and 5% withdrawal fee on PredictIt bets is known to distort markets."
```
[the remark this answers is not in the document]
anthonyebiner: Good point, I should mention that. Those along with the $850 contract betting cap decrease the efficiency of the markets. Long shot bets tend to be inflated, while safer bets with small returns are deflated. The 10% profit fee also adds an element of inefficiency markets with linked outcomes, inflating all of the prices.

notebooks/predictit_clean_sweep.ipynb
```diff
@@ -0,0 +1,380 @@
...
+        "colab_type": "text"
+      },
+      "source": [
+        "In this example, we use the PredictIt odds of democrats winning the house, senate, and presidency to make a prediction on whether they will sweep all three.\n",
+        "\n",
+        "PredictIt is composed of markets, which in turn are composed of questions. Although a PredictIt market has a literal question as its name, for consistency with Metaculus and Foretold, ergo considers each option in a market to be a separate \"question\". \n",
+        "\n",
+        "For instance, in the PredictIt market \"Which party will win the 2020 U.S. presidential election?\", one such question would be the democrat option. This is a binary question that essentially asks \"will democrats win the 2020 U.S. presidential election\".\n",
+        "\n",
+        "One thing to keep in mind is that the 10% profit fee, 5% withdrawal fee, and $850 betting cap are know to decrease the efficiency of markets. Long shot bets tend to be inflated while safer bets with lower margins are deflated. They also inflate prices in markets with linked outcomes, resulting in situations where adding up the prices of each contract comes to over $1."
```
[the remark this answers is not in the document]
anthonyebiner: Jesus I don't know how I'm missing all of these dumb grammar mistakes. I swear I read through the whole thing like twice to make sure there weren't any left. Maybe reading it out loud would help. And getting a good nights sleep haha.

brachbach:
@anthonyebiner I think you're right that Travis was failing because of missing credentials

I created [this branch](https://github.com/oughtinc/ergo/pull/413) just to be able to run Travis with the proper credentials, let's see how it goes

brachbach:
Travis is failing on that other PR, but I think just because of a linting issue. Clean that up and we'll try again?

(I think you might be able to push to that branch I created yourself? Not sure if you're authorized or not.)

ergo/platforms/predictit.py
```diff
@@ -0,0 +1,242 @@
...
+        self.market.refresh()
+        self._data = self.market.get_question(self.id)._data
+
+    def sample_community(self) -> bool:
+        """
+        Sample from the PredictIt community distribution (Bernoulli).
+
+        :return: true/false
+        """
+        return flip(self.get_community_prediction())
```
zmpster, line 97: @brachbach  I don't the best way to comment on this code generally without going to this PR, but this code just doesn't work. `self.get_community_prediction()` isn't implemented.

anthonyebiner: I believe in a previous comment it was decided to remove get_community_prediction since it wasn't standardized across platforms. I could add it back if you thought it could be helpful, otherwise sample_community should probably be removed since it's broken currently.

brachbach: Yeah I think it's best to just remove it. @anthonyebiner do you want to make a PR for that?

brachbach: Thanks @zmpster for pointing this out!

PR #406, opened 2020-08-19
Incorporate linting fixes from Elicit

brachbach:
Note that the Linear issue explains what this is/why this is important


brachbach:
@stuhlmueller can you review when you get a chance?

Issue #219, opened 2020-04-28
Make the logistic mixture fitting code more robust


stuhlmueller:

stuhlmueller:
Demonstrated in https://github.com/oughtinc/ergo/pull/220

stuhlmueller:
Remove `xfail` from tests once this issue is closed.

NixGD:
@stuhlmueller I'm curious for a status update on this.  It's something I've looked into a little bit and would be curious to try to get to the bottom of.  I'm not sure if there's a particular reason the issue was closed.

stuhlmueller:
I assume it works okay now. Do you know if there are tests or demos that show mixture fitting failing?

NixGD:
There are three tests that are set to xfail which all depend on the mixture-fitting. All three of them still fail (one also needed updating to the new interface).

I did a little bit of digging into why.  See [this branch](https://github.com/oughtinc/ergo/tree/demonstrate-fitting-failures) and especially [this notebook](https://github.com/oughtinc/ergo/blob/f30e8a97244ff3eb13f7973d852961aeed5e0de9/notebooks/Fitting%20failure%20demonstration.ipynb).

The short version is I think there are two problems:

1. Ergo implicitly clipping logistic distributions parameters to certain ranges, meaning that it fails to fit weirder distributions.
2. The function that prepares the distribution for metaculus submission assumes the distribution's scale is from 0 to 1, but it isn't always (even if the data is normalized for that range)

NixGD:
6a80b73 should fix point 2 above.

PR #416, opened 2020-09-14
Fix scaling for logistic fit

NixGD:
This branch fixes a problem where ergo would successfully fit a logistic mixture to a distribution but then skew it subtly.  Specifically, when called without passing a scale, the logistic mixture fitting function will use an appropriate scale based on the min and max of the data.  We were then fitting to the normalized data, but the resulting distribution had a scale that was not exactly `[0,1]`. The function that prepares the data for Metaculus submission assumes that it is given a normalized distribution, causing the submitted distribution to be shifted slightly.

This PR also includes a new test for fitting a fairly smooth distribution that the current master code fails.  Unfortunately, the existing fitting tests are still failing for other reasons.  Specifically, ergo seems to struggle to fit very spiky distributions; see #219 for some recent discussion of this.

### DIFF
--- a/ergo/platforms/metaculus/question/continuous.py
+++ b/ergo/platforms/metaculus/question/continuous.py
@@ -322,7 +322,10 @@ class ContinuousQuestion(MetaculusQuestion):
 
         normalized_samples = self.scale.normalize_points(samples)
         _dist = dist.LogisticMixture.from_samples(
-            normalized_samples, fixed_params={"num_components": 3}, verbose=verbose
+            normalized_samples,
+            fixed_params={"num_components": 3},
+            verbose=verbose,
+            scale=Scale(0, 1),
         )
         return self.prepare_logistic_mixture(_dist)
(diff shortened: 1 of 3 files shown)


brachbach (approved):
This looks good to me @NixGD!

Thanks for figuring out a test that actually passes :)

I tried removing your addition of `scale=Scale(0, 1),` from  `ergo/platforms/metaculus/question/continuous.py `, and the test failed!

That's cool, I wasn't expecting it to be sensitive enough to pick up the difference from the skew

brachbach:
@stuhlmueller you also have to approve this as code owner

like I wrote in my comment, I think this looks good

Our code here is a little crufty, e.g. the failing tests that @NixGD mentioned

But I think @NixGD did a great job of working around the crufty bits to write a great short PR that fixes this bug + tests the fix

stuhlmueller (approved):

PR #417, opened 2020-09-16
Fixes #408

### LINKED ISSUE #408: The show_prediction function for Metaculus questions raises a read-only error
The following code demonstrates the problem:
```python3
metaculus = ergo.Metaculus(username="oughtpublic", password="123456")
# generate any arbitary list of samples
samples = np.random.random(100)*100

question = metaculus.get_question(5053)
question.show_prediction(samples)
```

This gives the following error:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-35-63a8672c8161> in <module>
      5 
      6 question = metaculus.get_question(5053)
----> 7 question.show_prediction(samples)

~/Dropbox/Projects/Ergo/ergo/ergo/platforms/metaculus/question/continuous.py in show_prediction(self, samples, plot_samples, plot_fitted, percent_kept, side_cut_from, show_community, num_samples, **kwargs)
    488 
    489         for col in df:
--> 490             df[col] = self.scale.denormalize_points(df[col])
    491 
    492         df = pd.melt(df, var_name="sources", value_name="samples")  # type: ignore

~/Library/Caches/pypoetry/virtualenvs/ergo-1dtPmU33-py3.6/lib/python3.6/site-packages/pandas/core/frame.py in __setitem__(self, key, value)
   2936         else:
   2937             # set column
-> 2938             self._set_item(key, value)
   2939 
   2940     def _setitem_slice(self, key, value):

~/Library/Caches/pypoetry/virtualenvs/ergo-1dtPmU33-py3.6/lib/python3.6/site-packages/pandas/core/frame.py in _set_item(self, key, value)
   2999         self._ensure_valid_index(value)
   3000         value = self._sanitize_column(key, value)
-> 3001         NDFrame._set_item(self, key, value)
   3002 
   3003         # check if we are modifying a copy

~/Library/Caches/pypoetry/virtualenvs/ergo-1dtPmU33-py3.6/lib/python3.6/site-packages/pandas/core/generic.py in _set_item(self, key, value)
   3622 
   3623     def _set_item(self, key, value) -> None:
-> 3624         self._data.set(key, value)
   3625         self._clear_item_cache()
   3626 

~/Library/Caches/pypoetry/virtualenvs/ergo-1dtPmU33-py3.6/lib/python3.6/site-packages/pandas/core/internals/managers.py in set(self, item, value)
   1088             blk_locs = blklocs[val_locs.indexer]
   1089             if blk.should_store(value):
-> 1090                 blk.set(blk_locs, value_getitem(val_locs))
   1091             else:
   1092                 unfit_mgr_locs.append(blk.mgr_locs.as_array[blk_locs])

~/Library/Caches/pypoetry/virtualenvs/ergo-1dtPmU33-py3.6/lib/python3.6/site-packages/pandas/core/internals/blocks.py in set(self, locs, values)
    374         None
    375         """
--> 376         self.values[locs] = values
    377 
    378     def delete(self, loc):

ValueError: assignment destination is read-only

```
The same error occurs with other coninious Metaculus questions as well.  From a quick look it seems that these lines are enough to replicate the error:
```python3
df = pd.DataFrame()
df["samples"] = question.scale.normalize_points(samples)
df["samples"] = question.scale.denormalize_points(samples)
```

I believe this problem is at least partially due to normalize_points returning an immutable Jax DeviceArray object.

NixGD:
Fixes a read-only error resulting from data being an immutable `jax.np` array instead of a normal numpy array.  Also adds two tests that ensure `show_prediction()` runs without errors (although it does not test the accuracy of the graphs produced).

### DIFF
--- a/ergo/platforms/metaculus/question/continuous.py
+++ b/ergo/platforms/metaculus/question/continuous.py
@@ -464,9 +464,9 @@ class ContinuousQuestion(MetaculusQuestion):
                         "For multiple predictions comparisons, only samples can be compared (plot_fitted must be False)"
                     )
                 for col in samples:
-                    df[col] = self.scale.normalize_points(samples[col])
+                    df[col] = onp.array(self.scale.normalize_points(samples[col]))
             else:
-                df["samples"] = self.scale.normalize_points(samples)
+                df["samples"] = onp.array(self.scale.normalize_points(samples))
 
         if plot_fitted:
             prediction = self.get_submission_from_samples(samples)
--- a/ergo/platforms/metaculus/question/continuous.py
+++ b/ergo/platforms/metaculus/question/continuous.py
@@ -464,8 +464,10 @@ class ContinuousQuestion(MetaculusQuestion):
                         "For multiple predictions comparisons, only samples can be compared (plot_fitted must be False)"
                     )
                 for col in samples:
+                    # use numpy array to ensure df doesn't become read-only
                     df[col] = onp.array(self.scale.normalize_points(samples[col]))
             else:
+                # use numpy array to ensure df doesn't become read-only
                 df["samples"] = onp.array(self.scale.normalize_points(samples))
 
         if plot_fitted:
--- a/tests/test_metaculus.py
+++ b/tests/test_metaculus.py
@@ -3,9 +3,6 @@ from http import HTTPStatus
 import pprint
 
 import jax.numpy as np
-import matplotlib.pyplot as plt
-import numpy as onp
-import pandas as pd
 import pytest
 import requests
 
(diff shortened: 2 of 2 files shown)


ergo/platforms/metaculus/question/continuous.py
```diff
@@ -464,9 +464,9 @@ def show_prediction(
                         "For multiple predictions comparisons, only samples can be compared (plot_fitted must be False)"
                     )
                 for col in samples:
-                    df[col] = self.scale.normalize_points(samples[col])
+                    df[col] = onp.array(self.scale.normalize_points(samples[col]))
```
brachbach, line 467: @NixGD could you add a comment in the code here about what exactly this fixes?

(both for my benefit reviewing the PR, and for the benefit of future code maintainers)

tests/test_metaculus.py
```diff
@@ -212,3 +215,32 @@ def test_get_community_prediction_log(metaculus_questions):
 def test_sample_community_binary(metaculus_questions):
     value = metaculus_questions.binary_question.sample_community()
     assert bool(value) in (True, False)
+
+
+def test_show_prediction_with_fitted(
```
brachbach, line 220: I'm not a big fan of just testing that the code runs without erroring (as opposed to actually testing that it does the desired thing). I think we should just get rid of these tests

But if you feel like we should keep them, add docstrings explaining why

brachbach (changes requested):
Thanks for catching and fixing this, @NixGD!

Had a couple comments to address before we merge

tests/test_metaculus.py
brachbach: (TBC I am grateful that you included these for the sake of this PR, so we can see that #408 is indeed fixed, but I don't think it's worth keeping them)

NixGD:
Changes made as requested!  I don't understand the internals of pandas well enough to descirbe the problem in full detail; I can't find good documentation on what can and cannot be done to immutable data in pandas dataframes.

brachbach (approved):
Thanks for making the changes! I'll merge

Issue #89, opened 2020-04-17
Support use of Metaculus API without providing username and password


stuhlmueller:
Then update quickstart

stuhlmueller:
People should only need to enter their login information if they want to call methods that need it.

AABoyles:
Cheap way to accomplish this quickly: set the default argument values of `username` and `password` to the corresponding values for oughtpublic`. Bonus points for checking that the username on the session isn't oughtpublic before submitting predictions upstream.

zmpster:
I think this is done? `get_question` doesn't require `is_logged_in_via_uname_pwd`

PR #419, opened 2020-09-24
Allow for Metaculus authentication via API keys

uvafan:


NixGD:
Have you tested what happens if a user authenticated by an API key tries to call `get_question` while filtering by `player_status`?  Since `user_id` is never set I think it would return an empty dataframe as is, but maybe it should throw an error instead.

I wonder if it's also worth including #89 in this PR as well.  I've thought about what changes are needed to make that work and would be happy to try to incorporate them into here later today if that would be helpful (or I could work on them after this is merged).

ergo/platforms/metaculus/metaculus.py
```diff
@@ -67,12 +67,29 @@ class Metaculus:
         "interested": "upvoted_by",
     }
 
-    def __init__(self, username: str, password: str, api_domain: str = "www"):
+    def __init__(
+        self,
+        username: str = None,
+        password: str = None,
+        user_api_key: str = None,
+        org_api_key: str = None,
```
NixGD, line 75: The type hints for these should be Optional[str], right?

ergo/platforms/metaculus/metaculus.py
```diff
@@ -67,12 +67,29 @@ class Metaculus:
...
+        if username and password:
+            self.login(username, password)
+        elif user_api_key and org_api_key:
+            self.auth_method = "api_keys"
+        else:
+            raise ValueError(
+                "Authentication method not provided. Please provide either username and password, or user and org API keys."
+            )
 
     def login(self, username, password):
```
NixGD, line 94: What should happen when a user calls `.login()` on a Metaculus object that was set up with API authentication?

uvafan: Fair point. I think it would be reasonable to switch to user/pass authentication, I will implement this.

uvafan:
Thanks for the comments @NixGD! 

> Have you tested what happens if a user authenticated by an API key tries to call `get_question` while filtering by `player_status`? Since `user_id` is never set I think it would return an empty dataframe as is, but maybe it should throw an error instead.

I've updated the PR to allow for a `user_id` to be specified when using API key authentication, and to throw an error if it hasn't been specified but a filter which needs it is specified.

> I wonder if it's also worth including #89 in this PR as well. I've thought about what changes are needed to make that work and would be happy to try to incorporate them into here later today if that would be helpful (or I could work on them after this is merged).

 I'm going to try to merge these changes ASAP as they will make the Metaculus submission flow in Elicit cleaner, but you can base a PR for fixing #89 off of this one, or wait until it's merged.

brachbach (changes requested):
1. Tests are currently failing, I didn't look at why
2. I think you should add tests of the different login methods. I think for each of uname/pwd login and api key login:
    1. log in
    2. test to make sure a user-specific detail was properly set or can be retrieved. `user_id` seems like the most obvious for uname/pwd, not sure what makes the most sense for API keys
    2. check to make sure that you can access some user-specific detail from

brachbach:
Hmm on further thought, I think the structure of this class doesn't really make sense with 2 different forms of auth

If it weren't a breaking API change, I'd propose restructuring:

1. constructor just takes `api_domain`
2. after constructing, you can call either `login_with_username_and_password` or `login_with_api_key` (you can think of better names)
3. you should also be able to do most things without logging in, we should only require login where it's actually necessary
    1. but we can leave this for a PR that addresses #89 

given that it is a breaking API change, I think we should probably just leave it as-is

brachbach:
per Andreas on Slack, I think we should go ahead and make the change to the class structure

uvafan:
@brachbach, I pushed the requested change to the class structure, plus error handling to reflect that API key authentication can currently only be used for prediction. I also added a test for API key auth (there are existing tests like you described for user/pass auth).

ergo/platforms/metaculus/metaculus.py
```diff
@@ -67,14 +69,21 @@ class Metaculus:
...
 
-    def __init__(self, username: str, password: str, api_domain: str = "www"):
-        self.user_id = None
+    def __init__(
+        self,
+        api_domain: Optional[str] = "www",
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+    ):
+        if username and password:
```
brachbach, line 78: should be `or` I think

brachbach (approved):
Cool, looking much better!

Reflecting on it more, `auth_method` feels overly complicated to me, and the `post` method seems unnecessarily abstract (given that it's only used once in the codebase, to `.predict`). I think instead I'd:

1. get rid of `auth_method`
2. create `is_logged_in_via_uname_pwd()` and `have_api_keys()` methods
3. get rid of the `post` method
4. In `predict`:
    1. if `is_logged_in_via_uname_pwd()`, then use the uname/password auth. If not:
    2. if `have_api_keys()`, then use the API keys. If not:
    3. error

I don't feel that this is strictly necessary to do before merging, so I'll approve. But I think I'd suggest doing it before merging.

brachbach:
@uvafan I think this is TMI, but I also had this alternative, more abstract idea to get rid of `auth_method` and `post`:

1. don't store an `auth_method`. Instead, create `is_logged_in_via_uname_pwd()` and `have_api_keys()` methods.
2. change `post` to use whichever form of auth we have that allows us to do the most:
    1. if `is_logged_in_via_uname_pwd()`, then use the uname/password auth. If not:
    2. if `have_api_keys()`, then use the API keys. If not:
    3. try with no auth
3. if Metaculus returns `403 forbidden` for either the API keys or no auth, error to the user, "Metaculus forbids this action for the most permissive form of authentiation you currently have, which is []"

But, I think the simpler, concrete version I described in my review comment is better

Just including this in case you like it better for some reason

ergo/platforms/metaculus/metaculus.py
uvafan: Yeah I wasn't really sure which it should be, I'll just make it `or`, not sure it matters much.

NixGD:
I pushed some tests for unauthenticated users.  Note that one currently fails (`get_questions_json` checks for `user_id`, which may not be set)

uvafan:
@NixGD thanks for the tests! They should all pass now.

stuhlmueller (approved):

PR #428, opened 2020-11-02
Add '/' to Metacalculus questions url to remove redirect

### LINKED ISSUE #363: Remove unnecessary redirect in Metaculus API requests
Because / at end of question is missing

Screen Shot 2020-06-07 at 3 38 11 PM

zmpster:
Closes #363. There wasn't a great way to test this without exposing too much from the Metaculus class, thought if you disallow redirects, as I have done, but don't add the `/`, the tests will fail.

### DIFF
--- a/ergo/platforms/metaculus/metaculus.py
+++ b/ergo/platforms/metaculus/metaculus.py
@@ -186,7 +186,7 @@ class Metaculus:
         :param id: Question id (can be read off from URL)
         :param name: Name to assign to this question (used in models)
         """
-        r = self.s.get(f"{self.api_url}/questions/{id}")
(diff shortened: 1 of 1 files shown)


brachbach (changes requested):
Nice!

Actually my current guess is that we do still want to allow redirects (so that this code will still work if Metaculus ever changes their API URL in the future and adds a redirect to the new URL)

Do you think that makes sense @zmpster?

zmpster:
Yeah, that makes sense. I was thinking disallowing redirect explicitly was a bit too cute. Removed that option!

brachbach (approved):
Thanks for making that change!

I tested locally, everything works :)

merging

Issue #412, opened 2020-08-26
Support Python 3.7.* and 3.8.*


stuhlmueller:

stuhlmueller:
PR #411 demonstrates that 3.7 and 3.8 are not currently supported.

NixGD:
PR #414 is an attempt to fix this.

peterhurford:
Perhaps with https://github.com/oughtinc/ergo/pull/415 we can close this issue? Though we would need something like https://github.com/oughtinc/ergo/pull/430 to fix the error on installation

PR #429, opened 2020-11-11
Added a platform-independent questions-interface and integration with…

zmpster:
I've done two things in this PR:
1. I've added a questions directory and files to make a platform-independent questions interface that can be the basis for adding new platform integrations
2. I've added an integration with the Almanis prediction platform using the above interface (https://www.almanisprivate.com/almanis-app) with partial functionality. It allows you to fetch questions or a list of questions from the platform, get the current price/probability. This only works for binary questions currently.

For now I just wanted to get feedback before I go to deep on something that may not be helpful! This PR could be merged as is, and wouldn't cause any issues, but there are a few obvious things I could add:
1. Add in support for Almanis "multiple choice" questions which would map to something like Ergo's PointDensity distribution
2. Add a sample notebook using the integration
3. Add "submit" functionality so one could actually make predictions on the platform
4. Unify the other 3 platforms to use the new questions interface

If one or more of these things seems especially worthwhile to put in this PR, please let me know! If not, I'd be happy to get more fine-grained code feedback.

### DIFF
--- a/ergo/platforms/almanis.py
+++ b/ergo/platforms/almanis.py
@@ -1,0 +1,226 @@
+from abc import abstractmethod
+import json
+from typing import Dict, List, Sequence
+
+import requests
+
+from ergo.questions import BinaryQuestion, Question
+
+
+class Almanis:
+    """
+    The main class for interacting with Almanis
+    """
+
+    question_id_request_json = json.dumps(
+        {
+            "customerID": "44d7f0b3-7e6f-4867-b00a-6816492bb510",
+            "statuses": [{"Trading": {}}, {"ClosedForTrading": {}}],
+        }
+    )
+
+    BINARY_QUESTION_KEY = "BooleanQuestionOD"
+    CONTINUOUS_QUESTION_KEY = "MultipleChoiceQuestionOD"
+
+    def __init__(self):
+        self.api_url = "https://proddysruptlabsr53tobackendapi.dysruptlabs.com/api/sharedCode/api/IntegrationLayerApi"
+        self.s = requests.Session()
+        self.question_ids = []
+
+    def refresh_question_ids(self):
+        response = self.s.post(
+            f"{self.api_url}/getQuestionIDs", data=self.question_id_request_json
+        )
+        response_json = response.json()
+        self.question_ids = [x["id"] for x in response_json]
+
+    def get_questions(self, ids: List[str]) -> Sequence["AlmanisQuestion"]:
+        request_json = json.dumps({"questionIDs": ids})
+        questions_response = self.s.post(
+            f"{self.api_url}/getQuestionsWithIDs", data=request_json
+        )
+        questions_json = questions_response.json()
+
+        prices_response = self.s.post(
+            f"{self.api_url}/getCurrentPricesForQuestionIDs", data=request_json
+        )
+        prices_json = prices_response.json()
+        prices_by_id = {p["questionID"]: p for p in prices_json}
+        questions = []
+        for question_json in questions_json:
+            if self.BINARY_QUESTION_KEY in question_json:
+                question_data = question_json[self.BINARY_QUESTION_KEY]
+                binary_question = AlmanisBinaryQuestion(
+                    self, question_data, prices_by_id[question_data["id"]]
+                )
+                questions.append(binary_question)
+            else:
+                pass  # For now, ignore the multiple choice questions
+        return questions
+
+    def get_question(self, question_id: str) -> "AlmanisQuestion":
+        return self.get_questions([question_id])[0]
+
+    def get_all_questions(self) -> Sequence["AlmanisQuestion"]:
+        self.refresh_question_ids()
+        return self.get_questions(self.question_ids)
+
+
+class AlmanisQuestion(Question):
+    """
+    A question on the Almanis platform.
+
+    :param market: Almanis market instance
+    :param data: Contract JSON retrieved from PredictIt API
+
+    :ivar str id: id of the question
+    :ivar text: name of the question
+    :ivar status: status of the question
+
+    An Almanis question looks like:
+
+    {
+        "BooleanQuestionOD": {
+            "id": "5d4350e5-7721-43f6-901d-9c4f362b992c",
+            "customerID": "44d7f0b3-7e6f-4867-b00a-6816492bb510",
+            "shortName": null,
+            "text": "By 12 Dec 2020, will the S&P 500 fall below 2,800?",
+            "description": "<p style=\"text-align:start;\"><span style=\"color: rgba(0,0,0,0.87)..."
+            "settlementDescription": "",
+            "settledAnswer": null,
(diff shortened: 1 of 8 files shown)


ergo/questions/binary.py
```diff
@@ -0,0 +1,33 @@
...
+        """
+        Submit a prediction to the prediction platform
+
+        :param p: how likely is the event to happen, from 0 to 1?
+        """
+        raise NotImplementedError("This should be implemented by a subclass")
+
+    def sample_community(self) -> bool:
+        """
+        Sample from the PredictIt community distribution (Bernoulli).
```
zmpster, line 29: TODO: Remove "PredictIt" from comment

ergo/platforms/almanis.py
```diff
@@ -0,0 +1,226 @@
...
+    CONTINUOUS_QUESTION_KEY = "MultipleChoiceQuestionOD"
+
+    def __init__(self):
+        self.api_url = "https://proddysruptlabsr53tobackendapi.dysruptlabs.com/api/sharedCode/api/IntegrationLayerApi"
+        self.s = requests.Session()
+        self.question_ids = []
+
+    def refresh_question_ids(self):
+        response = self.s.post(
+            f"{self.api_url}/getQuestionIDs", data=self.question_id_request_json
```
uvafan, line 32: can use the `json=` argument rather than `json.dumps` then `data=`

ergo/platforms/almanis.py
```diff
@@ -0,0 +1,226 @@
...
+
+    question_id_request_json = json.dumps(
+        {
+            "customerID": "44d7f0b3-7e6f-4867-b00a-6816492bb510",
+            "statuses": [{"Trading": {}}, {"ClosedForTrading": {}}],
+        }
+    )
+
+    BINARY_QUESTION_KEY = "BooleanQuestionOD"
+    CONTINUOUS_QUESTION_KEY = "MultipleChoiceQuestionOD"
```
uvafan, line 23: Seems like there's a difference between multiple choice and continuous; multiple choice is >2 discrete choices, continuous is like any real value between 0 and 100.

uvafan: (Because of this, I also don't think it makes sense to try to use `PointDensity` to represent multiple-choice questions)

ergo/platforms/almanis.py
```diff
@@ -0,0 +1,226 @@
...
+        self.question_ids = []
+
+    def refresh_question_ids(self):
+        response = self.s.post(
+            f"{self.api_url}/getQuestionIDs", data=self.question_id_request_json
+        )
+        response_json = response.json()
+        self.question_ids = [x["id"] for x in response_json]
+
+    def get_questions(self, ids: List[str]) -> Sequence["AlmanisQuestion"]:
```
uvafan, line 37: Does including `"`s in the type hints work? I would think it should just be e.g. `Sequence[AlmanisQuestion]` but maybe both work

ergo/questions/binary.py
```diff
@@ -0,0 +1,33 @@
...
+class BinaryQuestion(Question):
+    @abstractmethod
+    def get_community_prediction(self) -> float:
+        """
+        Get the latest community probability for the binary event
+        """
+        raise NotImplementedError("This should be implemented by a subclass")
+
+    @abstractmethod
+    def submit(self, p: float, confidence: float = 0) -> requests.Response:
```
uvafan, line 19: What is the difference between `p` and `confidence`? Does it extend to all prediction platforms?

ergo/questions/question.py
```diff
@@ -0,0 +1,28 @@
+from abc import ABC, abstractmethod
+
+
+class Question(ABC):
+    """A question from a forecasting platform"""
+
+    @abstractmethod
+    def get_text(self):
```
uvafan, line 8: Not sure `get_text` is the best name here; maybe `get_description` or `get_title`?

uvafan (commented):
Overall thoughts:
- I think a question abstraction is a good idea; I'm less sure about the actual contents of the abstraction until we actually try to integrate it with other platforms
- Of your options, I think (1) doesn't really make sense as multiple choice isn't continuous; (2) and (3) seem like good next steps. (4) seems like a lot of work to get right but if you're up for it might be worth trying, would probably want to to first create a plan then get feedback on that before implementing.

ergo/questions/question.py
zmpster: I don't have a strong opinion here, but if we're talking about a "question", I feel like a normal English phrase would be "the question's text" and not the "question's title" or the "question's description". Happy to change either way though.

ergo/questions/binary.py
zmpster: Yeah, this was the abstraction I was least confident in (pun slightly intended). For both Almanis and PredictIt, you both submit a prediction to the platform, and then some number of points/dollars to "bet" on that prediction. That was the idea of confidence. This wouldn't apply for Foretold or Metaculus predictions.

ergo/platforms/almanis.py
zmpster: Yeah, I tried it without `"`s at first, but apparently Python type hinting doesn't work for a class you're declaring in the same file. You can get around this by using [postponed evaluation](https://www.python.org/dev/peps/pep-0563/#enabling-the-future-behavior-in-python-3-7), but that only works in Python 3.7 which it looks like Ergo doesn't support yet (https://github.com/oughtinc/ergo/issues/412)

ergo/platforms/almanis.py
zmpster: Oh, yeah, this is much better, thanks!

ergo/platforms/almanis.py
zmpster: Yeah, I mean in practice this representation is not "continuous" as it is a few concrete options, but I think the Almanis platform uses this type of question to ask something like "What will some continuous value be in the future" as opposed to "Will this thing happening, yes or no". For example an Almanis question (I can't link it because of how their web app works) asks about what the Dow will be at on 12/14, and [this Metaculus question](https://www.metaculus.com/questions/3472/what-will-teslas-market-capitalization-be-on-1-january-2021/) asks about Tesla's Market cap in 2021. Even though responses are represented differently, this seems like the same type of question to me. For example, in some world where the questions were phrased exactly the same (say both were asking about the Dow), we'd like to compare these two distributions as like objects. Am I making sense here?

Also, as I was poking around about this, there *are* some questions in Almanis that are true categorical multiple choice questions (e.g. "Which vaccine will be the most administered as at 12 June 2021?"), and those would obviously not qualify has the type of "continuous" question I'm talking about. I would have to make a new multiple choice class for those.

zmpster:
@uvafan Thanks for looking at this! I started to work on (4), though I had some specific questions on the existing Metaculus integration I may include for you and @brachbach.

brachbach:
Closing this one as well since we're no longer actively working on Ergo

PR #430, opened 2020-12-05
Allow pyproject.toml to use Python 3.7* and 3.8*

peterhurford:
Now that https://github.com/oughtinc/ergo/pull/415 is merged, can we permit Python 3.7* and 3.8* on install? This will avoid the `ERROR: Package 'ergo' requires a different Python: 3.8.6 not in '>=3.6.8,<3.7.0'` install error.

*WARNING: This PR has not been tested*

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ description = ""
 authors = ["Ought <team@ought.org>"]
 
 [tool.poetry.dependencies]
-python = "^3.6.8"
+python = "^3.8.5"
 requests = "2.21.0"
 country_converter = "0.6.7"
 tqdm = "4.45.0"


peterhurford:
It looks like there's an issue with Readthedocs

peterhurford:
Perhaps not helpful, but worth noting that this at least installs successfully on my laptop

pyproject.toml
```diff
@@ -5,7 +5,7 @@ description = ""
 authors = ["Ought <team@ought.org>"]
 
 [tool.poetry.dependencies]
-python = "^3.6.8"
+python = "^3.8.5"
```
brachbach, line 8: Hmmm... I'm pretty sure `python = "^3.8.5"` means ">= 3.8.5 <4.0.0": https://python-poetry.org/docs/versions/

According to https://python-poetry.org/docs/versions/, the current code should mean ">= 3.6.8 <4.0.0", which should work

I wonder if the `.python-version` file, or some other file, is causing the error message ERROR: Package 'ergo' requires a different Python: 3.8.6 not in '>=3.6.8,<3.7.0' (really not sure though)

brachbach (commented):
Re: CI failing, it's actually because the tests are failing

PRs from branches outside our Github org lack access to some secrets that the test need (I need to fix this but haven't gotten to it yet)

So for now, once you address my comment, I'll run CI with the secrets by creating my own copy of this branch

brachbach:
@peterhurford thanks for your work on this!

Unfortunately, since we're no longer actively working on Ergo, I'm going to close it.

PR #432, opened 2020-12-11
Fix Metaculus date parsing

uvafan:


brachbach (approved):
1. Would be nice to test this, but not sure how we would exactly and not a big deal
2. lame that the tests are erratically failing. Any idea why? My current hypothesis is that we're getting rate-limited by the Metaculus API because we're hitting it too often across the 6 different times that we run the tests (run tests twice [branch + PR] * run tests on 3 Python versions)
    1. TBC if this is accurate, there's nothing wrong with your PR, but any PR would run into the same problem

uvafan:
> 1. Would be nice to test this, but not sure how we would exactly and not a big deal
> 2. lame that the tests are erratically failing. Any idea why? My current hypothesis is that we're getting rate-limited by the Metaculus API because we're hitting it too often across the 6 different times that we run the tests (run tests twice [branch + PR] * run tests on 3 Python versions)
>    
>    1. TBC if this is accurate, there's nothing wrong with your PR, but any PR would run into the same problem

On 1, yeah would be nice but don't think it's worth the time for us to do right now.
On 2, yeah I'm also guessing that we're getting rate-limited.

PR #433, opened 2020-12-15
Almanis integration

brachbach:
Test if CI runs for https://github.com/oughtinc/ergo/pull/429


brachbach:
@zmpster thanks for your work on this!

However, since we're no longer actively working on Ergo, I'm going to go ahead and close it.

Sorry this didn't make it in!

PR #436, opened 2021-01-03
Rachbach/oug 1101 make ci run better

brachbach:
This is to test https://github.com/oughtinc/ergo/pull/435

to make sure the tests run properly for PRs made from a fork


brachbach:
This didn't work right off the bat, so closing now that we're no longer actively working on Ergo

